PR #186, opened 2020-08-05
Implement get the boundary value for the index scan

bright-starry-sky:
When the index scan, need to get the boundary value for where condition. for example : 
field c1 is int : 
- c1 > 5 --> (the range pair :  {max , c1 + 1}) 
- c1 > 1 and c1 < 5 --> (the range pair : {c1 - 1, c1 -1} )

field c1 is double , and the precise number of decimal places is 6 : 
- c1 > 0.5 --> (the range pair :  {max , 0.500001}) 
- c1 > 1 and c1 < 5 --> (the range pair : {1.000001, 4.999999} )


bright-starry-sky:
Close this PR. #263 will instead of this.

PR #207, opened 2020-08-12
Refactor expression access with visitor pattern

yixinglu:
depends on https://github.com/vesoft-inc-private/nebula-common/pull/194


dutor (commented):
Excellent work!

Please take care of the recent changes of expressions, renamed and newly-added ones.

PR #218, opened 2020-08-18
add boundclause tostring

nevermore3:


src/parser/TraverseSentences.cpp
```diff
@@ -222,4 +222,30 @@ std::string YieldSentence::toString() const {
...
+    buf.reserve(256);
+    buf += "GET SUBGRAPH ";
+    if (step_ != nullptr) {
+        buf += step_->toString();
+    }
+    if (from_ != nullptr) {
+        buf += " ";
+        buf += from_->toString();
+    }
+    if (in_ != nullptr) {
```
Shylock-Hg, line 237: There is the `in` , `out` and `both` information already.

src/parser/Clauses.h
```diff
@@ -357,8 +357,15 @@ class GroupClause final {
 
 class InBoundClause final {
 public:
-    explicit InBoundClause(OverEdges *edges) {
+    enum BoundType : uint8_t {
+        IN,
```
Shylock-Hg, line 361: It's extra.

src/parser/TraverseSentences.cpp
nevermore3: pls be more specific

src/parser/Clauses.h
```diff
@@ -367,10 +374,12 @@ class InBoundClause final {
 
 private:
     std::unique_ptr<OverEdges> overEdges_;
+    BoundType boundType_;
 };
 
-using OutBoundClause = InBoundClause;
-using BothInOutClause = InBoundClause;
+using InBoundClause = BoundClause;
```
Shylock-Hg, line 380: InBoundClause with ctor to set `IN` as value of `boundType_`.

laura-ding (approved):

yixinglu (approved):

PR #222, opened 2020-08-18
Fix join input.

CPWstatic:


src/validator/GoValidator.cpp
```diff
@@ -249,6 +247,7 @@ Status GoValidator::oneStep(PlanNode* dependencyForGn,
     } else {
         root_ = projectResult;
     }
+    VLOG(1) << root_->kind();
```
dutor, line 250: Some logs are for end users, while most are for developers. Please try logging with necessary contexts to make it easy to understand for other developers.

dutor (approved):

laura-ding (approved):

PR #223, opened 2020-08-18
expr tree rebuilt and visit

Shylock-Hg:
[NG-213] Support multi-visitors in once traversal.


yixinglu:
We have used visitor pattern to access expression tree and also could do what you want in single traversal. I think the code is more readable in that way. So we don't need this PR any more. I close it now, Thanks for your efforts.

PR #226, opened 2020-08-19
Add explain for some plan nodes.

Shylock-Hg:


src/planner/Admin.cpp
```diff
@@ -72,5 +72,94 @@ std::unique_ptr<cpp2::PlanNodeDescription> GetConfig::explain() const {
...
+std::unique_ptr<cpp2::PlanNodeDescription> DropUser::explain() const {
+    auto desc = DropNode::explain();
+    addDescription("username", *username_, desc.get());
+    return desc;
+}
+
+std::unique_ptr<cpp2::PlanNodeDescription> UpdateUser::explain() const {
+    auto desc = SingleDependencyNode::explain();
+    addDescription("username", *username_, desc.get());
+    addDescription("password", *password_, desc.get());
```
yixinglu, line 104: Shall we return password without any mask to console ?

yixinglu (commented):
Except password others LGTM

src/planner/Admin.cpp
Shylock-Hg: Maybe we could add note in document.

laura-ding: > Shall we return password without any mask to console ?

Agree, can't return the password directly. need to modify it.

src/planner/Admin.cpp
```diff
@@ -72,5 +72,94 @@ std::unique_ptr<cpp2::PlanNodeDescription> GetConfig::explain() const {
...
+    addDescription("username", *username_, desc.get());
+    addDescription("spaceName", *spaceName_, desc.get());
+    addDescription("role", meta::cpp2::_RoleType_VALUES_TO_NAMES.at(role_), desc.get());
+    return desc;
+}
+
+std::unique_ptr<cpp2::PlanNodeDescription> ChangePassword::explain() const {
+    auto desc = SingleDependencyNode::explain();
+    addDescription("username", *username_, desc.get());
+    addDescription("password", *password_, desc.get());
```
laura-ding, line 127: ditto

src/planner/Admin.cpp
```diff
@@ -72,5 +72,94 @@ std::unique_ptr<cpp2::PlanNodeDescription> GetConfig::explain() const {
...
+
+std::unique_ptr<cpp2::PlanNodeDescription> ListUserRoles::explain() const {
+    auto desc = SingleDependencyNode::explain();
+    addDescription("username", *username_, desc.get());
+    return desc;
+}
+
+std::unique_ptr<cpp2::PlanNodeDescription> ListRoles::explain() const {
+    auto desc = SingleDependencyNode::explain();
+    addDescription("space", util::toJson(space_), desc.get());
```
laura-ding, line 140: spaceId

src/planner/Admin.cpp
Shylock-Hg: This won't execute in fact, so the password will not stored.

Shylock-Hg: Had fixed.

laura-ding (approved):

yixinglu (approved):

PR #232, opened 2020-08-20
Set the get vertices/edges response columns name by plan. Change _vid…

Shylock-Hg:
… column to VertexID.

Change _vid column to VertexID.


CPWstatic (approved):

src/validator/GoValidator.cpp
```diff
@@ -683,6 +684,25 @@ std::vector<storage::cpp2::VertexProp> GoValidator::buildDstVertexProps() {
...
 
+std::vector<std::string> GoValidator::buildDstVertexColNames() {
+    std::vector<std::string> dstColNames;
+    dstColNames.reserve(exprProps_.dstTagProps().size());
+    dstColNames.emplace_back(kVid);
+    for (const auto &tagProps : exprProps_.dstTagProps()) {
+        auto tagId = tagProps.first;
+        auto tagNameResult = qctx_->schemaMng()->toTagName(space_.id, tagId);
+        DCHECK(tagNameResult.ok());
+        if (!tagNameResult.ok()) {
```
laura-ding, line 695: if `tagNameResult` is not ok, couldn't make validator continue. I think the tagName can be saved in exprProps_, so here not need to get tagName from tagId.

yixinglu (approved):
LGTM

src/validator/GoValidator.cpp
Shylock-Hg: This had be checked out of this functions, for more independent we could return error.

Shylock-Hg: But I don't think it's meanful.

Shylock-Hg: Fixed.

laura-ding (approved):

yixinglu (approved):

PR #238, opened 2020-08-24
Use GraphStatus

laura-ding:


src/util/GraphStatus.h
```diff
@@ -0,0 +1,546 @@
...
+
+private:
+    static cpp2::Language toLanguage(const std::string &l);
+
+    static cpp2::Encoding toEncoding(const std::string &e);
+
+    static std::string encode(const std::string &errorMsg);
+
+private:
+    ErrorCode        errorCode_{ErrorCode::SUCCEEDED};
```
Shylock-Hg, line 535: Why not use Status directly?

laura-ding: `Status` is an internal error, driven by error message, and cannot be used to enforce the error code returned. if enforce it use error code, internal use is not convenient.

Shylock-Hg: I don't think add a new one it better.

Shylock-Hg: You can cast internal error when return to user and still is a Status.

laura-ding: Cann’t do that. You end up with an internal error, and status shouldn't depend on Thrift.

Shylock-Hg: Why shouldn't depend on thrift error code?

Shylock-Hg:
I think we need to use one enum to represent all error in the whole nebula system, and could know error come from which component by prefix number.
And we need provide the error message template to standardize the error message.

laura-ding:
> I think we need to use one enum to represent all error in the whole nebula system, and could know error come from which component by prefix number.
> And we need provide the error message template for each error to standardize the error message.

Now it is using the same enum, Please have a look at the common repo.

Shylock-Hg:
> > I think we need to use one enum to represent all error in the whole nebula system, and could know error come from which component by prefix number.
> > And we need provide the error message template for each error to standardize the error message.
> 
> Now it is using the same enum, Please have a look at the common repo.

ok

src/util/GraphStatus.h
yixinglu: I also think it's convenient to use single status definition.

PR #239, opened 2020-08-24
Add AttributeExpression

dutor:
The syntax counterpart of vesoft-inc/nebula-common#207


CPWstatic (approved):

yixinglu (approved):
Well Done

src/parser/parser.yy
```diff
@@ -400,46 +397,143 @@ base_expression
...
+        $$ = new LabelExpression($1);
+    }
+    | INTEGER {
+        $$ = new ConstantExpression($1);
+    }
+    | compound_expression {
         $$ = $1;
     }
-    | src_ref_expression {
+    | MINUS {
```
Shylock-Hg, line 409: Will this make `-` a valid expression in parser?

dutor: > Will this make `-` a valid expression in parser?

Actually it is a mid-rule action after the keyword `MINUS`

Shylock-Hg (approved):

laura-ding (approved):

PR #240, opened 2020-08-25
Let plan node depend on QueryContext rather than ExecutionPlan

yixinglu:
When rewriting execution plan, we don't know which plan new plan node belongs to.


Shylock-Hg:
Why need this?

yixinglu:
> Why need this?

the purpose has been described in this pr description.

CPWstatic (approved):

laura-ding (approved):

PR #241, opened 2020-08-25
Pull dependencies fields Up to plan node class

yixinglu:
This PR will simplify the usage of dependencies when don't care about the plan node kind.

In addition, fix logical executors include guard macros.

### DIFF
--- a/src/planner/PlanNode.h
+++ b/src/planner/PlanNode.h
@@ -23,6 +23,12 @@ class PlanNodeDescription;
  */
 class PlanNode {
 public:
+    enum class DepKind : uint8_t {
+        kSingleDep = 0,
+        kBinaryDep,
+        kNoDep,
+    };
+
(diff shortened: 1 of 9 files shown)


src/planner/PlanNode.h
```diff
@@ -198,11 +209,11 @@ class SingleInputNode : public SingleDependencyNode {
 class BiInputNode : public PlanNode {
 public:
     void setLeft(const PlanNode* left) {
-        left_ = left;
+        setDep(0, left);
```
laura-ding, line 212: 0, 1 use const var to replace

yixinglu: I think it's not necessary to define two const variables to represent left/right positions, it's simple enough to understand.

CPWstatic (approved):

laura-ding (approved):

PR #242, opened 2020-08-26
Support to save default value to expression

laura-ding:
depends on vesoft-inc/nebula-common#211

### DIFF
--- a/src/util/SchemaUtil.cpp
+++ b/src/util/SchemaUtil.cpp
@@ -202,8 +202,8 @@ StatusOr<DataSet> SchemaUtil::toDescSchema(const meta::cpp2::Schema &schema) {
                 break;
             }
             if (expr->kind() == Expression::Kind::kConstant) {
-                QueryExpressionContext ctx(nullptr, nullptr);
-                defaultValue = Expression::eval(expr.get(), ctx);
+                QueryExpressionContext ctx;
+                defaultValue = Expression::eval(expr.get(), ctx(nullptr));
             } else {
                 defaultValue = Value(expr->toString());
             }
@@ -247,8 +247,8 @@ StatusOr<DataSet> SchemaUtil::toShowCreateSchema(bool isTag,
                 break;
             }
             if (expr->kind() == Expression::Kind::kConstant) {
-                QueryExpressionContext ctx(nullptr, nullptr);
-                auto& value = expr->eval(ctx);
+                QueryExpressionContext ctx;
+                auto& value = expr->eval(ctx(nullptr));
                 auto toStr = value.toString();
                 if (value.isNumeric() || value.isBool()) {
                     createStr += " DEFAULT " + toStr;
--- a/src/util/SchemaUtil.cpp
+++ b/src/util/SchemaUtil.cpp
@@ -199,7 +199,7 @@ StatusOr<DataSet> SchemaUtil::toDescSchema(const meta::cpp2::Schema &schema) {
             if (expr == nullptr) {
                 LOG(ERROR) << "Internal error: Wrong default value expression.";
                 defaultValue = Value();
-                break;
+                continue;
             }
             if (expr->kind() == Expression::Kind::kConstant) {
                 QueryExpressionContext ctx;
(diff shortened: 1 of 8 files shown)


src/validator/MaintainValidator.cpp
```diff
@@ -16,6 +17,51 @@
...
+        column.set_name(*spec->name());
+        column.set_type(type);
+        column.set_nullable(spec->isNull());
+        if (meta::cpp2::PropertyType::FIXED_STRING == type) {
+            column.set_type_length(spec->typeLen());
+        }
+
+        if (spec->hasDefaultValue()) {
+            if (!evaluableExpr(spec->getDefaultValue())) {
+                return Status::SemanticError("Wrong default value experssion `%s'",
```
Shylock-Hg, line 41: More clear error message , such as `Constant expression required`

src/validator/MaintainValidator.cpp
```diff
@@ -16,6 +17,51 @@
...
+            auto defaultValueExpr = spec->getDefaultValue();
+            auto& value = defaultValueExpr->eval(ctx);
+            auto valStatus = SchemaUtil::toSchemaValue(type, value);
+            if (!valStatus.ok()) {
+                return valStatus.status();
+            }
+            // When the timestamp value is string, need to save the int value
+            if (type == meta::cpp2::PropertyType::TIMESTAMP && value.isStr()) {
+                ConstantExpression newExpr(std::move(valStatus).value());
+                column.set_default_value(newExpr.encode());
```
Shylock-Hg, line 54: If you store the ConstantExpression it will produce the default value when insert new data.

src/validator/MaintainValidator.cpp
laura-ding: `Constant expression` is the internal name, we just tell the user where it error

src/util/SchemaUtil.cpp
```diff
@@ -224,7 +193,21 @@ StatusOr<DataSet> SchemaUtil::toDescSchema(const meta::cpp2::Schema &schema) {
...
         auto nullable = col.__isset.nullable ? *col.get_nullable() : false;
         row.values.emplace_back(nullable ? "YES" : "NO");
-        auto defaultValue = col.__isset.default_value ? *col.get_default_value() : Value();
+        auto defaultValue = Value::kEmpty;
+        if (col.__isset.default_value) {
+            auto expr = Expression::decode(*col.get_default_value());
+            if (expr == nullptr) {
+                LOG(ERROR) << "Internal error: Wrong default value expression.";
+                defaultValue = Value();
+                break;
```
yixinglu, line 202: continue ?

src/util/SchemaUtil.cpp
```diff
@@ -257,12 +240,23 @@ StatusOr<DataSet> SchemaUtil::toShowCreateSchema(bool isTag,
...
         if (col.__isset.default_value) {
-            auto value = col.get_default_value();
-            auto toStr = value->toString();
-            if (value->isNumeric() || value->isBool()) {
-                createStr += " DEFAULT " + toStr;
+            auto encodeStr = *col.get_default_value();
+            auto expr = Expression::decode(encodeStr);
+            if (expr == nullptr) {
+                LOG(ERROR) << "Internal error: the default value is wrong expression.";
+                break;
```
yixinglu, line 247: ditto

src/util/ToJson.cpp
```diff
@@ -80,7 +83,9 @@ folly::dynamic toJson(const meta::cpp2::ColumnDef &column) {
         obj.insert("nullable", folly::to<std::string>(*column.get_nullable()));
     }
     if (column.__isset.default_value) {
-        obj.insert("defaultValue", column.get_default_value()->toString());
+        graph::QueryExpressionContext ctx;
+        auto value = Expression::decode(*column.get_default_value())->eval(ctx(nullptr));
+        obj.insert("defaultValue", value.toString());
```
yixinglu, line 88: It's not necessary to update here. as you say, we should response to user the `now()` string rather than current timestamp.

CPWstatic (approved):

src/util/SchemaUtil.cpp
laura-ding: good catch

src/util/ToJson.cpp
laura-ding: good catch

yixinglu (approved):
LGTM

bright-starry-sky (approved):
LGTM.

PR #244, opened 2020-08-27
Add lifetime mark for result.

Shylock-Hg:


Shylock-Hg:
The lifetime implemented in #764 , the move semantic need more thinks.

PR #247, opened 2020-08-27
Add optimizer 

yixinglu:
depends on #240 

This PR has not consider cost calculation of each plan node which has equal default cost 0.0. And I will continue to implement more better strategy to select best plan. 

By given a `PushFilterDownGetNbrsRule` example, others could add more rules like that and test the rule by comparing the optimized plan with expected result in `test_optimizer.py`.

This's one tiny step for optimization of Graph Query engine. More works are to be continued.

### DIFF
--- a/src/optimizer/Optimizer.cpp
+++ b/src/optimizer/Optimizer.cpp
@@ -1,0 +1,106 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "optimizer/Optimizer.h"
+
+#include "context/QueryContext.h"
+#include "optimizer/OptGroup.h"
+#include "optimizer/OptRule.h"
+#include "planner/ExecutionPlan.h"
+#include "planner/Logic.h"
+#include "planner/PlanNode.h"
+
+using nebula::graph::BiInputNode;
+using nebula::graph::Loop;
+using nebula::graph::PlanNode;
+using nebula::graph::QueryContext;
+using nebula::graph::Select;
+using nebula::graph::SingleDependencyNode;
+
+namespace nebula {
+namespace opt {
+
+Optimizer::Optimizer(QueryContext *qctx, const RuleSet *ruleSet) : qctx_(qctx), ruleSet_(ruleSet) {
+    DCHECK(qctx != nullptr);
+    DCHECK(ruleSet != nullptr);
+}
+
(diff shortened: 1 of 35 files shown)


src/optimizer/Optimizer.cpp
```diff
@@ -0,0 +1,106 @@
...
+Status Optimizer::doExploration() {
+    // TODO(yee): Apply all rules recursively, not only once round
+    for (auto rule : ruleSet_->rules()) {
+        if (!rootGroup_->isExplored(rule)) {
+            NG_RETURN_IF_ERROR(rootGroup_->explore(rule));
+        }
+    }
+    return Status::OK();
+}
+
```
bright-starry-sky, line 54: How to set the default rule when `enable_optimizer=false`?

yixinglu: I'll control rule set selection in query instance. pls wait a minute.

yixinglu: Now you can add opt rule to `RuleSet::defaultRules()` like [PushFilterDownGetNbrsRule](https://github.com/vesoft-inc/nebula-graph/pull/247/files#diff-1628cc036fa9a70c831e51ce2cb0884fR32)

src/optimizer/rule/PushFilterDownGetNbrsRule.h
```diff
@@ -0,0 +1,43 @@
...
+namespace opt {
+
+class PushFilterDownGetNbrsRule final : public OptRule {
+public:
+    static std::unique_ptr<OptRule> kInstance;
+
+    bool match(const OptGroupExpr *groupExpr) const override;
+    Status transform(graph::QueryContext *qctx,
+                     const OptGroupExpr *groupExpr,
+                     TransformResult *result) const override;
```
bright-starry-sky, line 28: A question, Way `const` of method transform? Doing so will not allow set private variables in class xxxxRule. WDYT?

dutor (approved):
LGTM.

bright-starry-sky:
Please resolve the conflict. Thanks.

src/optimizer/rule/PushFilterDownGetNbrsRule.h
yixinglu: Each rule will be used by many queries, and we register each rule in the singleton way into rule set. so change state of each rule will affect optimization of other queries. we should keep each rule stateless.

yixinglu:
@bright-starry-sky @dutor resolved conflicts.

bright-starry-sky (approved):
LGTM.

dutor (approved):
LGTM.

PR #252, opened 2020-09-01
Add operator() for expression context.

CPWstatic:


yixinglu (approved):
It's more clear for evaluation of expression applied on each value pointed by iterator.

laura-ding (approved):

PR #259, opened 2020-09-03
Add rewrite label attr expr visitor

yixinglu:
Refactor expression util implementation with expr visitor.


dutor (commented):
IMO, here the __Find__ and __Collect__ visitor could be abstracted to two more high level kinds: FindAnyVistor and CollectAllVisitor, which you construct with a predicate.

BTW. there are several names better to be renamed, like:
 * findAnyKind -> findAny
 * findAllKindInAll -> collectAll
 * hasAnyKind -> hasAny

src/visitor/FindExprVisitor.h
```diff
@@ -0,0 +1,71 @@
...
+    void visit(VertexExpression* expr) override;
+    void visit(EdgeExpression* expr) override;
+
+    void visitBinaryExpr(BinaryExpression* expr) override;
+
+    void findExpr(const Expression* expr);
+
+    bool found_{false};
+    const Expression* expr_{nullptr};
+    const std::unordered_set<Expression::Kind>& exprs_;
```
dutor, line 65: `kinds_`

yixinglu:
> IMO, here the __Find__ and __Collect__ visitor could be abstracted to two more high level kinds: FindAnyVistor and CollectAllVisitor, which you construct with a predicate.
> 
> BTW. there are several names better to be renamed, like:
>  * findAnyKind -> findAny
>  * findAllKindInAll -> collectAll
>  * hasAnyKind -> hasAny

good idea！i will update this pr according to your advice again.

src/visitor/FindAnyExprVisitor.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (found_) return;
+    expr->operand()->accept(this);
+}
+
+void FindAnyExprVisitor::visit(FunctionCallExpression *expr) {
+    findExpr(expr);
+    if (found_) return;
+    for (const auto &arg : expr->args()->args()) {
+        arg->accept(this);
+        if (found_) break;
```
laura-ding, line 34: better to use `return`, keep the same as under code.

src/visitor/CollectAllExprsVisitor.h
```diff
@@ -0,0 +1,67 @@
...
+#include "visitor/ExprVisitorImpl.h"
+
+namespace nebula {
+namespace graph {
+
+class CollectAllExprsVisitor final : public ExprVisitorImpl {
+public:
+    explicit CollectAllExprsVisitor(const std::unordered_set<Expression::Kind>& exprKinds);
+    bool ok() const override {
+        return true;
```
laura-ding, line 22: return !exprs_.empty();

src/visitor/FindAnyExprVisitor.cpp
yixinglu: fixed

src/visitor/CollectAllExprsVisitor.h
yixinglu: here should always return TRUE for traversing all nodes in expression tree.

src/visitor/FindExprVisitor.h
yixinglu: fixed

laura-ding (approved):

yixinglu:
@dutor @laura-ding resolved conflicts. Please review this PR again.

laura-ding (approved):

dutor (approved):
Bravo!

PR #260, opened 2020-09-03
Fix failure when all field are constant expressions in yield sentence with filter

laura-ding:


src/validator/YieldValidator.cpp
```diff
@@ -162,6 +179,11 @@ Status YieldValidator::validateYieldAndBuildOutputs(const YieldClause *clause) {
 
         NG_RETURN_IF_ERROR(makeOutputColumn(column->clone().release()));
     }
+
+    if (!hasInputVar) {
+        // generate constant expression result into querycontext
+        genConstantExprValues();
+    }
```
yixinglu, line 186: I think we maybe need to check whether `yield->where()` expression is a constant expression, if so we can control the filter node what rows to output. Here `hasInputVar` only checks yield columns, not include where condition expression. please consider this situation: `yield 1 as id | yield 'hello' where $-.id > 0`

tests/query/v1/test_yield.py
```diff
@@ -581,3 +581,14 @@ def test_map(self):
...
+    def test_constant_expression_with_where(self):
+        query = 'YIELD 1+1, "Hello world!" WHERE true;'
+        resp = self.execute_query(query)
+        self.check_resp_succeeded(resp)
+        self.check_result(resp, [[2, 'Hello world!']])
+
+        query = 'YIELD 1+1, "Hello world!" WHERE NULL;'
+        resp = self.execute_query(query)
+        self.check_resp_succeeded(resp)
+        self.check_empty_result(resp)
```
yixinglu, line 594: Very good test cases. Thx.

src/validator/YieldValidator.cpp
laura-ding: good catch

yixinglu (approved):
LGTM

dutor (approved):

PR #261, opened 2020-09-07
Add QueryStorageExecutor to extract the code to handle completeness f…

Shylock-Hg:
…rom storage responses and the failed partition.


dutor (approved):
Well done.

src/executor/query/QueryStorageExecutor.h
```diff
@@ -4,46 +4,42 @@
...
 #include "executor/Executor.h"
+#include "common/clients/storage/StorageClientBase.h"
 
 namespace nebula {
 namespace graph {
 
-class MutateExecutor : public Executor {
-public:
-    MutateExecutor(const std::string &name, const PlanNode *node, QueryContext *qctx)
+// It's used for data write/update/query
```
yixinglu, line 16: It's better to move this file to `executor/` rather than `executor/query` since it's used by mutate/query executors.

Shylock-Hg: Fixed, thanks.

yixinglu (approved):
LGTM

dutor (approved):

PR #263, opened 2020-09-07
Bound value for index

bright-starry-sky:
When the index scan, need to get the boundary value for where condition. for example :
field c1 is int :

- c1 > 5 --> (the range pair : {max , c1 + 1})
- c1 > 1 and c1 < 5 --> (the range pair : {c1 - 1, c1 -1} )

field c1 is double , and the precise number of decimal places is 6 :
- c1 > 0.5 --> (the range pair : {max , 0.500001})
- c1 > 1 and c1 < 5 --> (the range pair : {1.000001, 4.999999} )


src/optimizer/OptimizerUtils.cpp
```diff
@@ -0,0 +1,334 @@
...
+            return v;
+        }
+        case Value::Type::STRING : {
+            if (!col.__isset.type_length || col.get_type_length() == nullptr) {
+                return v;
+            }
+            std::string str;
+            str.reserve(*col.get_type_length());
+            str.append(v.getStr());
+            str.append(*col.get_type_length() - v.getStr().size(), '\0');
```
dutor, line 177: The actual length of the value might be greater than expected.

src/optimizer/OptimizerUtils.cpp
```diff
@@ -0,0 +1,334 @@
...
+            std::vector<unsigned char> bytes(str.begin(), str.end());
+            for (size_t i = bytes.size(); i >= 1; i--) {
+                if (bytes[i-1] > 0) {
+                    bytes[i-1] -= 1;
+                    break;
+                }
```
dutor, line 183: `if !value.getStr().empty()`, why not:
```cpp
if (bytes[i-1]-- != 0) break; 
```

src/optimizer/OptimizerUtils.cpp
```diff
@@ -0,0 +1,334 @@
...
+                    d.month = 12;
+                    if (d.year > 1) {
+                        d.year -= 1;
+                    } else {
+                        return v.getDate();
+                    }
+                }
+            }
+            return Value(d);
+        }
```
dutor, line 208: I have the same doubt as above

src/optimizer/OptimizerUtils.cpp
bright-starry-sky: Good, Bright eyes.

bright-starry-sky:
Addressed @dutor's comments. Thanks .

src/optimizer/OptimizerUtils.cpp
```diff
@@ -0,0 +1,347 @@
...
+                str.append(*col.get_type_length() - v.getStr().size(), '\0');
+            }
+            std::string maxStr(*col.get_type_length(), '\377');
+            if (str == maxStr) {
+                return Value(maxStr);
+            }
+            std::vector<unsigned char> bytes(str.begin(), str.end());
+            for (size_t i = bytes.size(); i >= 1; i--) {
+                if (bytes[i-1]++ != 255) break;
+            }
```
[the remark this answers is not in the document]
dutor: Besides, should we regard a fixed string longer than expected as a error?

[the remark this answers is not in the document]
bright-starry-sky: > I think these could be improved a little bit, like:
> 
> ```c++
> std::vector<unsigned char> bytes(v.getStr().begin(), v.getStr().end());
> bytes.resize(*col.get_type_length());
> size_t i = bytes.size();
> for (; i > 0; i--) {
>     if (bytes[i-1]++ != 255) break;
> } 
> if (i == 0) {
>     return Value(std::string('\377', *col.get_type_length()));
> }
> return Value(std::string(bytes.begin(), bytes.end()));
> ```

learned, more clear.

bright-starry-sky:
Simplified code logic by @dutor's proposal.

dutor (approved):
LGTM.

yixinglu (approved):
LGTM

PR #265, opened 2020-09-08
Move show sentence into admin

darionyaphet:


src/parser/AdminSentences.cpp
```diff
@@ -24,6 +24,46 @@ std::string ShowPartsSentence::toString() const {
     return std::string("SHOW PARTS");
 }
 
+std::string ShowTagsSentence::toString() const {
```
laura-ding, line 27: Schema is belonged to maintain

darionyaphet: Actually I want to move `show sentence` into one place.

laura-ding: The classification is consistent with the validate and executor. I don’t think it needs to move.

darionyaphet: Emmm ... The core of this problem is that what do you think about `show` sentence. 

IF you think it is a admin sentence, you should put them together. 

According to you, `ShowUsersSentence` should put into `UserSentence`

laura-ding: Actual the `ShowUsersSentence` also belongs to the `admin`. Everything related to the user belongs to `Admin`.

yixinglu:
Now plan nodes are classified as following:
**admin**:
- user
- role
- space
- host
- snapshot

**maintain**:
- tag
- edge
- index

I agree with @laura-ding, if we need to change this classified method later, we should also update validator and executor related codes. 

Now i close the PR, If we need it, reopen it later. Thanks for your efforts!

PR #273, opened 2020-09-10
change null value to empty value

nevermore3:
depend on https://github.com/vesoft-inc/nebula-storage/pull/137

### DIFF
--- a/src/context/Iterator.cpp
+++ b/src/context/Iterator.cpp
@@ -191,7 +191,7 @@ const Value& GetNeighborsIter::getColumn(const std::string& col) const {
     auto& index = dsIndices_[segment].colIndices;
     auto found = index.find(col);
     if (found == index.end()) {
-        return Value::kNullValue;
+        return Value::kEmpty;
     }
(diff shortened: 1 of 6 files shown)


src/context/Iterator.cpp
```diff
@@ -206,17 +206,17 @@ const Value& GetNeighborsIter::getTagProp(const std::string& tag,
...
     if (propIndex == index->second.propIndices.end()) {
-        return Value::kNullValue;
+        return Value::kEmpty;
     }
     auto colId = index->second.colIdx;
     auto& row = *(iter_->row_);
     DCHECK_GT(row.size(), colId);
     if (!row[colId].isList()) {
-        return Value::kNullBadType;
+        return Value::kEmpty;
```
laura-ding, line 219: Value::kNullBadType

CPWstatic (dismissed):

src/context/Iterator.cpp
nevermore3: done

dutor (approved):
LGTM.

PR #275, opened 2020-09-10
add Top-N PlanNode and Executor

jievince:
NG-120
1. add Top-N PlanNode and Executor
2. support JoinIter and PropIter in SortExecutor and TopNExecutor
2. change the type of factors_ in Sort node to std::vector<std::pair<size_t, OrderFactor::OrderType>> to simply code.

### DIFF
--- a/src/executor/query/SortExecutor.cpp
+++ b/src/executor/query/SortExecutor.cpp
@@ -24,23 +24,23 @@ folly::Future<Status> SortExecutor::execute() {
         LOG(ERROR) << errMsg;
         return Status::Error(errMsg);
     }
-    if (iter->isSequentialIter()) {
-        auto seqIter = static_cast<SequentialIter*>(iter.get());
+    if (iter->isSequentialIter() || iter->isJoinIter() || iter->isPropIter()) {
+        // auto seqIter = static_cast<SequentialIter*>(iter.get());
         auto &factors = sort->factors();
-        auto &colIndices = seqIter->getColIndices();
-        std::vector<std::pair<size_t, OrderFactor::OrderType>> indexes;
-        for (auto &factor : factors) {
-            auto indexFind = colIndices.find(factor.first);
-            if (indexFind == colIndices.end()) {
-                LOG(ERROR) << "Column name `" << factor.first
-                           << "' does not exist.";
-                return Status::Error("Column name `%s' does not exist.",
-                                     factor.first.c_str());
-            }
-            indexes.emplace_back(std::make_pair(indexFind->second, factor.second));
-        }
-        auto comparator = [&indexes] (const LogicalRow &lhs, const LogicalRow &rhs) {
-            for (auto &item : indexes) {
+        // auto &colIndices = seqIter->getColIndices();
+        // std::vector<std::pair<size_t, OrderFactor::OrderType>> indexes;
+        // for (auto &factor : factors) {
+        //     auto indexFind = colIndices.find(factor.first);
+        //     if (indexFind == colIndices.end()) {
+        //         LOG(ERROR) << "Column name `" << factor.first
+        //                    << "' does not exist.";
+        //         return Status::Error("Column name `%s' does not exist.",
+        //                              factor.first.c_str());
+        //     }
+        //     indexes.emplace_back(std::make_pair(indexFind->second, factor.second));
+        // }
+        auto comparator = [&factors] (const LogicalRow &lhs, const LogicalRow &rhs) {
+            for (auto &item : factors) {
                 auto index = item.first;
                 auto orderType = item.second;
                 if (lhs[index] == rhs[index]) {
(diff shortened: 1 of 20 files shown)


src/executor/query/TopNExecutor.cpp
```diff
@@ -0,0 +1,101 @@
...
+            heapSize = size;
+        }
+        if (heapSize == 0) {
+            seqIter->clear();
+            return finish(ResultBuilder().value(iter->valuePtr()).iter(std::move(iter)).finish());
+        }
+
+        std::vector<SequentialIter::SeqLogicalRow> heap(seqIter->begin(),
+                                                        seqIter->begin() + heapSize);
+        std::make_heap(heap.begin(), heap.end(), comparator);
```
Shylock-Hg, line 79: Good job. But I think you could package these logical to one structure like `BinaryHeap`.

src/executor/query/TopNExecutor.cpp
```diff
@@ -0,0 +1,101 @@
...
+        while (it != seqIter->end()) {
+            if (comparator(*it, heap[0])) {
+                std::pop_heap(heap.begin(), heap.end(), comparator);
+                heap.pop_back();
+                heap.push_back(*it);
+                std::push_heap(heap.begin(), heap.end(), comparator);
+            }
+            ++it;
+        }
+        std::sort_heap(heap.begin(), heap.end(), comparator);
```
Shylock-Hg, line 90: I think the TopN don't require ordered.

jievince: TopN is an optimization for sort + limit, so TopN =sort+limit. So I think Top N also requires order.

Shylock-Hg: Ok, maybe the name TopN need show the ordered property like `OrderedTopN`

src/executor/query/TopNExecutor.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto iter = ectx_->getResult(topn->inputVar()).iter();
+    if (UNLIKELY(iter == nullptr)) {
+        return Status::Error("Internal error: nullptr iterator in topn executor");
+    }
+    if (UNLIKELY(iter->isGetNeighborsIter())) {
+        std::string errMsg = "Internal error: TopN executor does not supported GetNeighborsIter";
+        LOG(ERROR) << errMsg;
+        return Status::Error(errMsg);
+    }
+    if (iter->isSequentialIter()) {
```
CPWstatic, line 27: `JoinIter` is also acceptable.

src/executor/query/TopNExecutor.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+        auto offset = topn->offset();
+        auto count = topn->count();
+        auto size = seqIter->size();
+        int64_t maxCount = count;
+        int64_t heapSize = 0;
+        if (size <= static_cast<size_t>(offset)) {
+            maxCount = 0;
+        } else if (size > static_cast<size_t>(offset + count)) {
+            heapSize = offset + count;
```
yixinglu, line 67: I'm confused about the heap size is not count.

jievince: When the `offset` of Limit is 0, TopN executor gets the top `count` rows of data;
and when `offset` is not 0, TopN firstly gets the top `offset+count` rows of data, and then retrive `count` rows starting from `offset`. This is useful when offset is not 0 but closely to 0.
Because currently we can't know the amout of data in execution plan period, so the TopN optimization rule is just applied when offset is 0.

yixinglu: So it's meaningful only in the situation that offset is 0. In my opinion, TopN should not contains `offset`, it's different from Limit here.

jievince: I think topn=sort+limit.
When `offset` of `limit` is not 0 but closely to 0, eg. 1 or 2 or3, it may also useful to apply topn rule.
If we support CBO in the future, we may know the approximate amount of data in the optimization period so that we can evaluate whether it is necessary to apply topn rule.
For example, if data is about 10000 rows, and `offset` of `limit` is 3, it's obviously useful to apply topn rule, even though the `offset` is not 0. And if `offset` is 9990, it's obivously not usefule to apply topn rule.

src/executor/query/SortExecutor.cpp
```diff
@@ -19,45 +19,45 @@ folly::Future<Status> SortExecutor::execute() {
...
+    if (iter->isSequentialIter()) {
+        auto seqIter = static_cast<SequentialIter*>(iter.get());
         std::sort(seqIter->begin(), seqIter->end(), comparator);
+    } else if (iter->isJoinIter()) {
+        auto joinIter = static_cast<JoinIter*>(iter.get());
+        std::sort(joinIter->begin(), joinIter->end(), comparator);
+    } else if (iter->isPropIter()) {
+        auto propIter = static_cast<PropIter*>(iter.get());
+        std::sort(propIter->begin(), propIter->end(), comparator);
     }
```
CPWstatic, line 60:
```
std::sort(iter->begin(), iter->end(), comparator);
```
We don't have to do the cast.

CPWstatic (commented):
LGTM.

src/executor/query/SortExecutor.cpp
jievince: Class Iterator doesn't has begin() and end() methods.

yixinglu (approved):

CPWstatic (approved):

PR #278, opened 2020-09-11
fix bug in PushFilterDownGetNbrsRule

jievince:


src/optimizer/rule/PushFilterDownGetNbrsRule.cpp
```diff
@@ -81,6 +83,7 @@ Status PushFilterDownGetNbrsRule::transform(QueryContext *qctx,
         // Filter(A&&B)->GetNeighbors(C) => Filter(A)->GetNeighbors(B&&C)
         auto newGroup = OptGroup::create(qctx);
         newGroupExpr = OptGroupExpr::create(qctx, newGN, newGroup);
+        newGN->setOutputVar(gn->outputVar());
```
laura-ding, line 86: you need to move it into `cloneGetNbrs`

jievince: > you need to move it into `cloneGetNbrs`

No, it can't be moved into `cloneGetNbrs`

>     } else {
        // Filter(A)->GetNeighbors(C) => GetNeighbors(A&&C)
        newGroupExpr = OptGroupExpr::create(qctx, newGN, groupExpr->group());
        newGN->setOutputVar(filter->outputVar());
    }
in this case, newGn's outputVar is filter->outputVar(), not gn->outputVar()

yixinglu (dismissed):
Good Catch. Thx.

src/optimizer/rule/PushFilterDownGetNbrsRule.cpp
yixinglu: @jievince  you can do it like ding says since the output variable will be overrided if it's needed.

laura-ding (approved):

yixinglu (approved):

PR #279, opened 2020-09-11
Optimizer rule for TopN

jievince:


src/optimizer/rule/TopNRule.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (!pair.first) {
+        return false;
+    }
+
+    return true;
+}
+
+Status TopNRule::transform(QueryContext *qctx,
+                                            const OptGroupExpr *groupExpr,
+                                            TransformResult *result) const {
```
yixinglu, line 47: fix indent

src/executor/query/TopNExecutor.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+        auto offset = topn->offset();
+        auto count = topn->count();
+        auto size = seqIter->size();
+        int64_t maxCount = count;
+        int64_t heapSize = 0;
+        if (size <= static_cast<size_t>(offset)) {
+            maxCount = 0;
+        } else if (size > static_cast<size_t>(offset + count)) {
+            heapSize = offset + count;
```
yixinglu, line 67: I don't understand why heap size is `offset+count` rather than `count`.

yixinglu (approved):

yixinglu (approved):
Well Done.

yixinglu (approved):

laura-ding (approved):

yixinglu (approved):

Issue #56, opened 2020-05-22
Support client reconnect


laura-ding:
Now the interface of query engine can't support reconnect from the client, because the query engine not associating an existing session with a reconnected client.

PR #280, opened 2020-09-14
Support sessions saving by metad

laura-ding:
depends on https://github.com/vesoft-inc/nebula-common/pull/223

closed https://github.com/vesoft-inc/nebula-graph/issues/56


conf/nebula-graphd.conf.default
```diff
@@ -30,6 +30,8 @@
 --local_ip=127.0.0.1
 # Network device to listen on
 --listen_netdev=any
+# Local ip
+--local_ip=127.0.0.1
```
yixinglu, line 34: duplicate with line 30

src/executor/CMakeLists.txt
```diff
@@ -47,12 +47,15 @@ nebula_add_library(
...
+    admin/ConfigExecutor.cpp
+    admin/SessionExecutor.cpp
     maintain/TagExecutor.cpp
     maintain/EdgeExecutor.cpp
     mutate/InsertExecutor.cpp
     mutate/DeleteExecutor.cpp
     mutate/UpdateExecutor.cpp
-    admin/ConfigExecutor.cpp
+
+
```
yixinglu, line 58: cleanup

src/parser/Sentence.h
```diff
@@ -26,96 +27,15 @@ class Sentence {
...
-        kGroupBy,
-        kReturn,
-        kCreateSnapshot,
-        kDropSnapshot,
-        kAdminJob,
-        kGetSubgraph,
-    };
-
-    Kind kind() const {
+    SentenceKind kind() const {
```
yixinglu, line 30: where is the sentenceKind defined?

laura-ding: common repo

src/validator/test/ValidatorTestBase.h
```diff
@@ -110,7 +113,7 @@ class ValidatorTestBase : public ::testing::Test {
     static void bfsTraverse(const PlanNode* root, std::vector<PlanNode::Kind>& result);
 
 protected:
-    std::shared_ptr<Session>              session_;
+    std::shared_ptr<ClientSession>              session_;
```
darionyaphet, line 116: extra space?

src/parser/Sentence.h
darionyaphet: So move Sentence::Kind to common's SentenceKind ?

yixinglu: It's not convenient when we add new sentence like `Match`, so i think it's not a good practice to update thrift interfaces frequently。

src/session/ClientSession.h
```diff
@@ -23,16 +23,20 @@ struct SpaceInfo {
     meta::cpp2::SpaceDesc spaceDesc;
 };
 
-class Session final {
+class ClientSession final {
```
Shylock-Hg, line 26: Why rename?

laura-ding: common has the session struct to save the session infomation.

src/service/GraphService.cpp
```diff
@@ -17,39 +17,57 @@
...
+    // check username and password failed
+    if (!auth(username, password)) {
+        onHandle(*ctx, cpp2::ErrorCode::E_BAD_USERNAME_PASSWORD);
+        ctx->finish();
+        return ctx->future();
     }
 
-    ctx.finish();
-    return ctx.future();
+    auto future = ctx->future();
```
dutor, line 68: Acquire the future once after the creation of the ctx.

src/service/GraphService.cpp
```diff
@@ -65,18 +83,20 @@ GraphService::future_execute(int64_t sessionId, const std::string& query) {
...
-        }
-        ctx->setSession(std::move(result).value());
+    auto session = sessionManager_->findSessionFromCache(sessionId);
+    if (session != nullptr) {
+        session->updateGraphAddr(myAddr_);
+        ctx->setSession(std::move(session));
+        queryEngine_->execute(std::move(ctx));
+    } else {
+        auto cb = [this](std::unique_ptr<RequestContext<cpp2::ExecutionResponse>> rctx){
+            queryEngine_->execute(std::move(rctx));
```
dutor, line 93: What if the session not found at the meta side.

src/daemons/GraphDaemon.cpp
```diff
@@ -84,15 +84,9 @@ int main(int argc, char *argv[]) {
...
-    std::string localIP;
-    {
-        auto result = NetworkUtils::getIPv4FromDevice(FLAGS_listen_netdev);
-        if (!result.ok()) {
-            LOG(ERROR) << result.status();
-            return EXIT_FAILURE;
-        }
-        localIP = std::move(result).value();
-    }
+    auto hostIdentity = FLAGS_local_ip == "" ? NetworkUtils::getHostname()
```
dutor, line 87: Why use the hostname, instead of finding a usable IP?

dutor (commented):
Well done.

I think the GET SESSION and SHOW SESSION[S] could be united to one statement, as well as the related interfaces of meta

laura-ding:
> I think the GET SESSION and SHOW SESSION[S] could be united to a single statement, as well as the related interfaces of meta

the interface in meta is two, and I think to get session and list session better to use a different interface because the list session includes big data.

src/session/ClientSession.h
Shylock-Hg: It's not conflict.

dutor:
> > I think the GET SESSION and SHOW SESSION[S] could be united to a single statement, as well as the related interfaces of meta
> 
> the interface in meta is two, and I think to get session and list session better to use a different interface because the list session includes big data.

Well, you could keep the two interfaces while using a single statement at the frontend.

src/session/ClientSession.h
laura-ding: It's not about conflict, it's about separation

Shylock-Hg: It's done by the namespace already.

dutor (approved):
Well done.

dutor:
Please resolve the conflicts.

tests/Makefile
```diff
@@ -43,13 +43,14 @@ check:
...
-	PYTHONPATH=$$PYTHONPATH:$(CURR_DIR)/.. $(CURR_DIR)/nebula-test-run.py --cmd=start --rm_dir=$(RM_DIR) --build_dir=$(BUILD_DIR)
+	PYTHONPATH=$$PYTHONPATH:$(CURR_DIR)/.. $(CURR_DIR)/nebula-test-run.py --cmd=start --rm_dir=$(RM_DIR) --build_dir=$(BUILD_DIR) --multi_graphd=true
 
 down:
 	PYTHONPATH=$$PYTHONPATH:$(CURR_DIR)/.. $(CURR_DIR)/nebula-test-run.py --cmd=stop --rm_dir=$(RM_DIR)
 
 test:
-	cd $(CURR_DIR) && python3 -m pytest -n$(J) --dist=loadfile --rm_dir=$(RM_DIR) -m "not skip" -k "not tck" $(TEST_DIR)
+	cd $(CURR_DIR) && python3 -m pytest -n8 --dist=loadfile --rm_dir=$(RM_DIR) -m "not skip" $(TEST_DIR) \
+	&& python3 -m pytest -n8 --dist=loadfile --rm_dir=$(RM_DIR) -m "not skip" job/test_session.py
```
yixinglu, line 53: Add `job/test_session.py` to `python_files` section in `tests/ptest.ini`

tests/job/test_session.py
```diff
@@ -0,0 +1,183 @@
...
+                                         get_connection(addr_host2, addr_port2),
+                                         session_id,
+                                         i)
+                test_jobs.append(future)
+
+            for future in concurrent.futures.as_completed(test_jobs):
+                if future.exception() is not None:
+                    assert False, future.exception()
+                else:
+                    assert True
```
yixinglu, line 177: `assert future.exception() is None, future.exception()`

src/session/ClientSession.h
```diff
@@ -0,0 +1,149 @@
...
+
+    explicit ClientSession(meta::cpp2::Session &&session, meta::MetaClient* metaClient);
+
+
+private:
+    SpaceInfo               space_;
+    time::Duration          idleDuration_;
+    meta::cpp2::Session     session_;
+    meta::MetaClient*       metaClient_{ nullptr};
+    folly::RWSpinLock       rwSpinLock_;
```
yixinglu, line 137: There's some execution error if client runs different queries asynchronously. And it's better to mange client session in SesssionManager.

We should make sure that server doesn't run different queries of same session in parallel.

laura-ding: Yes, this problem has always existed, and this problem will be solved by another PR.

src/session/SessionManager.h
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        ~AuthInstance() = default;
+
+        RequestContext<AuthResponse>* rctx() const {
+            return rctx_.get();
+        }
+
+        void finish() {
+            delete this;
```
CPWstatic, line 87: I think `AuthInstance` and `ExecuteInstance` are over designed. They only wraps a `RequestContext` but nothing else todo. 
```
        void finish() {
            delete this;
        }
```
And this is not a good practice i think, a right value reference is enough for your scenarios. It will easily lead to a double free problem, since the `finish` is a public interface.
```
AuthInstance auth;
auth->finish();
```

src/session/SessionManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+void SessionManager::createSession(const std::string &userName,
+                                   const std::string& clientIp,
+                                   folly::Executor* runner,
+                                   std::unique_ptr<RequestContext<AuthResponse>> rctx) {
+    {
+        folly::RWSpinLock::ReadHolder rHolder(rwlock_);
+        if (activeSessions_.size() >= FLAGS_max_allowed_connections) {
+            return rctx->finish();
```
CPWstatic, line 154: I think the `SessionManager` has no responsibility to finish the `RequestContext`. And we should return error here.

src/session/SessionManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+                sessionPtr->charge();
+                activeSessions_[sid] = sessionPtr;
+                return sessionPtr;
+            }
+            return findPtr->second;
+        }
+    };
+
+    auto authInstance = new AuthInstance(std::move(rctx));
+    auto doFinish = [this,
```
CPWstatic, line 182: Again, i think it's the callers' duty to finish the `RequestContext`.

src/session/ClientSession.h
CPWstatic, line 137: Not quite sure how much impact the `lock` has on performance. Need a benchmark, i think.

CPWstatic: I'll do the benchmark.

src/daemons/GraphDaemon.cpp
```diff
@@ -89,15 +89,12 @@ int main(int argc, char *argv[]) {
...
     // Get the IPv4 address the server will listen on
-    std::string localIP;
-    {
-        auto result = NetworkUtils::getIPv4FromDevice(FLAGS_listen_netdev);
-        if (!result.ok()) {
-            LOG(ERROR) << result.status();
-            return EXIT_FAILURE;
-        }
-        localIP = std::move(result).value();
+    if (FLAGS_local_ip.empty()) {
```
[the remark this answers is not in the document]
laura-ding: The  local_ip is not allowed empty, Because it's stored in Metad, it is meaningless when empty.

src/session/SessionManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+            execInstance->rctx()->resp().errorCode = ErrorCode::E_SESSION_INVALID;
+            execInstance->rctx()->resp().errorMsg.reset(new std::string("Session not found."));
+            execInstance->rctx()->finish();
+            return execInstance->finish();
+        }
+        auto session = std::move(resp).value();
+        session->updateGraphAddr(myAddr_);
+        auto roles = metaClient_->getRolesByUserFromCache(session->user());
+        for (const auto &role : roles) {
+            session->setRole(role.get_space_id(), role.get_role_type());
```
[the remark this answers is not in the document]
laura-ding: This only happened in Graphd Crash

src/session/SessionManager.cpp
laura-ding: good catch

src/session/SessionManager.h
laura-ding: They are private classes and only managed by the SessionManager, which guarantees them.

CPWstatic: Did you miss the `over designed` part?

src/daemons/GraphDaemon.cpp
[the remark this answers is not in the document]
CPWstatic: I mean why change the logic of `getIPv4FromDevice` to this new one.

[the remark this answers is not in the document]
laura-ding: The `getIPv4FromDevice` will get the 0.0.0.0 ip.

src/session/SessionManager.cpp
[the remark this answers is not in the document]
CPWstatic: I don't get it, why it only happens when Graphd Crashes? My point is that `setRole ` is an expensive interface which locks every time.

CPWstatic:
close #1057

CPWstatic (commented):
LGTM.

src/daemons/GraphDaemon.cpp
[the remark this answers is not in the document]
Shylock-Hg: Does this related to session?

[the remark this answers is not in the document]
laura-ding: the session info will save the graph address, but it's not a direct relationship. It is only because local_ip must be set, otherwise the identities of different GraphDs may become indistinguishable on the machine.

CPWstatic (approved):
LGTM.

yixinglu (approved):

PR #284, opened 2020-09-14
Fix get edge.

CPWstatic:
After https://github.com/vesoft-inc/nebula-common/pull/227

### DIFF
--- a/src/context/Iterator.cpp
+++ b/src/context/Iterator.cpp
@@ -299,30 +299,25 @@ Value GetNeighborsIter::getEdge() const {
     if (!type.isInt()) {
         return Value::kNullBadType;
     }
+    edge.type = type.getInt();
 
     auto& src = getColumn(kVid);
     if (!SchemaUtil::isValidVid(src)) {
         return Value::kNullBadType;
     }
+    edge.src = src.getStr();
 
     auto& dst = getEdgeProp(edgeName, kDst);
     if (!SchemaUtil::isValidVid(dst)) {
         return Value::kNullBadType;
     }
-    if (type.getInt() > 0) {
-        edge.src = src.getStr();
-        edge.dst = dst.getStr();
-    } else {
-        edge.src = dst.getStr();
-        edge.dst = src.getStr();
-    }
+    edge.dst = dst.getStr();
 
     auto& rank = getEdgeProp(edgeName, kRank);
     if (!rank.isInt()) {
         return Value::kNullBadType;
     }
     edge.ranking = rank.getInt();
-    edge.type = 0;
 
     auto& edgePropMap = dsIndices_[segment].edgePropsMap;
     auto edgeProp = edgePropMap.find(currentEdgeName());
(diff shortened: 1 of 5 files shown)


dutor (dismissed):

nevermore3 (approved):

src/context/Iterator.h
```diff
@@ -273,8 +274,13 @@ class GetNeighborsIter final : public Iterator {
     List getEdges() {
         DCHECK(iter_ == logicalRows_.begin());
         List edges;
+        edges.values.reserve(size());
         for (; valid(); next()) {
-            edges.values.emplace_back(getEdge());
+            auto edge = getEdge();
+            if (edge.isEdge()) {
```
laura-ding, line 280: ditto

src/context/Iterator.cpp
```diff
@@ -579,7 +574,12 @@ List PropIter::getVertices() {
 List PropIter::getEdges() {
     DCHECK(iter_ == rows_.begin());
     List edges;
+    edges.values.reserve(size());
     for (; valid(); next()) {
+        auto edge = getEdge();
+        if (edge.isEdge()) {
```
laura-ding, line 580: The if in here is superfluous

CPWstatic: It could be null.

src/context/Iterator.cpp
```diff
@@ -579,7 +574,12 @@ List PropIter::getVertices() {
...
+            const_cast<Edge&>(edge.getEdge()).format();
+        }
         edges.values.emplace_back(getEdge());
```
laura-ding, line 583: edges.values.emplace_back(std::move(edge));
and put it into if

src/context/Iterator.h
```diff
@@ -273,8 +274,13 @@ class GetNeighborsIter final : public Iterator {
...
+                const_cast<Edge&>(edge.getEdge()).format();
+            }
+            edges.values.emplace_back(std::move(edge));
```
laura-ding, line 283: ditto

src/context/Iterator.cpp
laura-ding: ok

src/context/Iterator.cpp
CPWstatic: Null is also acceptable.

laura-ding (approved):

src/context/Iterator.cpp
yixinglu: I think what ding means is that you need not to call `getEdge()` again since `edge` has been there.

tests/common/nebula_test_suite.py
```diff
@@ -289,14 +294,16 @@ def value_to_string(self, value):
             return self.date_time_to_string(value.get_dtVal())
         elif value.getType() == CommonTtypes.Value.MVAL:
             return self.map_to_string(value.get_mVal())
-        return 'Unsupported type'
+        else:
+            return value.__repr__();
+        return 'Unsupported type: {}'.format(value.getType())
```
yixinglu, line 299: will never reach here since above else branch.

laura-ding (approved):

dutor (approved):

PR #285, opened 2020-09-14
Optimization rule for limit push down

jievince:
NG-239


src/optimizer/rule/LimitPushDownRule.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if (!gnFound) {
+        return std::make_pair(false, matched);
+    }
+
+    const auto gn = static_cast<const GetNeighbors *>(gnExpr->node());
+    if (limitRows >= gn->limit()) {
+        return std::make_pair(false, matched);
+    }
+    matched.push_back(projExpr);
+    matched.push_back(gnExpr);
```
yixinglu, line 135: Returned OptGroupExprs only contains the root GroupExpr rather than all touched GroupExprs

jievince: Why? If just return projExpr, function LimitPushDownRule::transform will execute the following code again:

>     bool gnFound = false;
    const OptGroupExpr *gnExpr = nullptr;
    for (auto dep : projExpr->dependencies()) {
        for (auto expr : dep->groupExprs()) {
            if (expr->node()->kind() == PlanNode::Kind::kGetNeighbors) {
                gnFound = true;
                gnExpr = expr;
                break;
            }
        }
    }

yixinglu: make sense. I'm thinking about whether need to collect these OptGroupExprs by patten match.

yixinglu (approved):
I think this PR could be merged at present, I will update it when OptRule interface changed.

src/optimizer/rule/LimitPushDownRule.cpp
```diff
@@ -0,0 +1,200 @@
...
+        newGn->setStatProps(std::move(statPropsPtr));
+    }
+
+    if (gn->exprs()) {
+        auto exprs = *gn->exprs();
+        auto exprsPtr = std::make_unique<decltype(exprs)>(std::move(exprs));
+        newGn->setExprs(std::move(exprsPtr));
+    }
+    return newGn;
+}
```
CPWstatic, line 197: Emm, these `clone`s should be provided by the node itself.

CPWstatic (dismissed):

jievince:
clone function is fixed in yee's pr https://github.com/vesoft-inc/nebula-graph/pull/311/files

src/optimizer/rule/LimitPushDownRule.cpp
jievince: this is fixed in yee's pr https://github.com/vesoft-inc/nebula-graph/pull/311/files

yixinglu (dismissed):
Good job!

src/optimizer/rule/LimitPushDownRule.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto projExpr = matched.dependencies.front().node;
+    auto gnExpr = matched.dependencies.front().dependencies.front().node;
+
+    const auto limit = static_cast<const Limit *>(limitExpr->node());
+    const auto proj = static_cast<const Project *>(projExpr->node());
+    const auto gn = static_cast<const GetNeighbors *>(gnExpr->node());
+
+    int64_t limitRows = limit->offset() + limit->count();
+    if (limitRows >= gn->limit()) {
+        return TransformResult{false, false, {}};
```
CPWstatic, line 57: It's difficult to understand, suggest add some static method for such situations, such as
```
TransformResult::emptyResult()
Or 
TransformResult::noTrans()
```

src/optimizer/OptRule.h
```diff
@@ -50,6 +50,10 @@ class OptRule {
         bool eraseCurr{false};
         bool eraseAll{false};
         std::vector<OptGroupExpr *> newGroupExprs;
+        static const TransformResult &noTransform() {
+            static TransformResult kNoTrans{false, false, {}};
+            return kNoTrans;
+        }
```
CPWstatic, line 56:
```
static method -> method -> static variable -> variable
```
Emm, the static method should be place at the beginning of the class/struct.

jievince: Thanks

yixinglu (approved):

yixinglu (dismissed):

CPWstatic (dismissed):

laura-ding (approved):

yixinglu (approved):

PR #288, opened 2020-09-15
Add constant expression folding visitor

yixinglu:
depends on vesoft-inc/nebula-common#230

### DIFF
--- a/src/validator/YieldValidator.cpp
+++ b/src/validator/YieldValidator.cpp
@@ -11,6 +11,7 @@
 #include "parser/Clauses.h"
 #include "parser/TraverseSentences.h"
 #include "planner/Query.h"
+#include "visitor/FoldConstantExprVisitor.h"
 
 namespace nebula {
 namespace graph {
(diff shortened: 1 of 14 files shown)


nevermore3 (approved):

src/validator/YieldValidator.cpp
```diff
@@ -192,6 +201,15 @@ Status YieldValidator::validateWhere(const WhereClause *clause) {
...
         NG_RETURN_IF_ERROR(deduceProps(filter, exprProps_));
+        auto newFilter = filter->clone();
+        FoldConstantExprVisitor visitor;
+        newFilter->accept(&visitor);
+        if (visitor.canBeFolded()) {
+            filterCondition_ = visitor.fold(newFilter.get());
+        } else {
+            filterCondition_ = newFilter.release();
+        }
+        qctx_->objPool()->add(filterCondition_);
```
dutor, line 212: IMO. this is tedious and error-prone to use the visitor like this. Could we wrap these into a function, e.g. `OptimizerUtils::foldConstantExpr(Expression*)`

yixinglu: I think putting this util function in ExpressionUtils is better than OptimizerUtils. Have fixed it.

dutor (approved):
Great work

laura-ding (approved):

PR #289, opened 2020-09-16
Remove offline flag when rebuild index

darionyaphet:
depend on [232](https://github.com/vesoft-inc/nebula-common/pull/232)


yixinglu (approved):

bright-starry-sky (approved):
LGTM

PR #291, opened 2020-09-18
Make static empty result initialize lazily

yixinglu:
Now you can add log info when build result.

Fix TIME property type compiler error introduced in vesoft-inc/nebula-common#239


src/parser/ColumnTypeDef.h
```diff
@@ -14,31 +14,15 @@ struct ColumnTypeDef {
...
-                return meta::cpp2::_PropertyType_VALUES_TO_NAMES.at(this->type);
-            case meta::cpp2::PropertyType::FIXED_STRING: {
-                std::string str;
-                str += meta::cpp2::_PropertyType_VALUES_TO_NAMES.at(this->type);
-                str += "(";
-                str += folly::to<std::string>(this->typeLen);
-                str += ")";
-                return str;
-            }
+        if (type == meta::cpp2::PropertyType::FIXED_STRING) {
```
laura-ding, line 17: I think it can use SchemaUtil::typeToString directly.

laura-ding (dismissed):

src/context/Result.h
```diff
@@ -21,15 +21,15 @@ class ResultBuilder;
...
-    static const std::vector<Result> kEmptyResultList;
-
     enum class State : uint8_t {
         kUnExecuted,
         kPartialSuccess,
         kSuccess,
     };
 
+    static const Result& emptyResult();
+    static const std::vector<Result>& emptyResultList();
```
dutor, line 31: To comply with `Status::OK()`, better to capitalize the first character.

yixinglu: Thx, Fixed.

laura-ding (approved):

jievince (approved):

PR #294, opened 2020-09-21
Update vid type definition in SpaceDesc

yixinglu:
depends on vesoft-inc/nebula-common#243, vesoft-inc/nebula-storage#147

close #298


laura-ding (dismissed):

dutor (approved):
Thanks for fixing these.

laura-ding (approved):

PR #295, opened 2020-09-22
Move the ExpressionProps out of validator folder.

Shylock-Hg:


laura-ding (dismissed):

laura-ding (approved):

src/visitor/DeducePropsVisitor.h
```diff
@@ -18,7 +18,67 @@ class Expression;
 namespace graph {
 
 class QueryContext;
-class ExpressionProps;
+
+class ExpressionProps final {
```
yixinglu, line 22: I prefer to move the ExpressionProps to visitor directory rather than merge it with visitor into single file.

Shylock-Hg: It's related to this visitor only.

nevermore3 (approved):

PR #301, opened 2020-09-23
Add pattern match for OptRule

yixinglu:
When we add new optimize rules, we must find expected OptGroupExpr twice in match and transform interfaces. This PR try to simplify this process by introducing the pattern match method.


CPWstatic (dismissed):

bright-starry-sky (approved):
LGTM.

CPWstatic (approved):

nevermore3 (approved):

PR #305, opened 2020-09-25
Fix yield $-.prop

jievince:
In the follow ngql, variable `srcAndEdgePropCols_` will not be initialized, but function `buildProjectSrcEdgePropsForGN` used this `srcAndEdgePropCols_`(nullptr) later 
```sql
"GO 1 STEPS FROM "1" OVER like YIELD like._dst AS\
      id | GO 1 STEPS FROM $-.id OVER like YIELD $-.id"
```


src/validator/test/QueryValidatorTest.cpp
```diff
@@ -229,6 +229,22 @@ TEST_F(QueryValidatorTest, GoWithPipe) {
         };
         EXPECT_TRUE(checkResult(query, expected));
     }
+    {
+        std::string query = "GO 1 STEPS FROM \"1\" OVER like YIELD like._dst AS "
+                            "id | GO 1 STEPS FROM $-.id OVER like YIELD $-.id";
```
Shylock-Hg, line 234: The second sentence will call buildProjectSrcEdgePropsForGN?

jievince: yes，the second go sentence will call this func，and it will cause graphd to crash
`if (!exprProps_.inputProps().empty() || !exprProps_.varProps().empty() ||
        !exprProps_.dstTagProps().empty() || from_.fromType != FromType::kInstantExpr) {
        projectSrcEdgeProps = buildProjectSrcEdgePropsForGN(gn->outputVar(), gn);
    }`

src/validator/GoValidator.cpp
```diff
@@ -949,8 +949,9 @@ Status GoValidator::buildColumns() {
     }
 
     auto pool = qctx_->objPool();
-    if (!exprProps_.srcTagProps().empty() || !exprProps_.edgeProps().empty() ||
-        !exprProps_.dstTagProps().empty()) {
+    if (!exprProps_.inputProps().empty() || !exprProps_.varProps().empty() ||
+        !exprProps_.srcTagProps().empty() || !exprProps_.edgeProps().empty() ||
+        !exprProps_.dstTagProps().empty() || from_.fromType != FromType::kInstantExpr) {
```
yixinglu, line 954: Extract these conditions to a function with a meaningful name will be better, like `isPropsEmpty()` or something ?

src/validator/test/QueryValidatorTest.cpp
yixinglu: Add test case in CI makes sense than comparing plan.

yixinglu (approved):

yixinglu (approved):

yixinglu (dismissed):

src/validator/test/QueryValidatorTest.cpp
Shylock-Hg: The plan test case could remain.

jievince: > The plan test case could remain.

There are already similar test cases in plan test

src/validator/GoValidator.cpp
```diff
@@ -798,5 +797,11 @@ Status GoValidator::buildColumns() {
     return Status::OK();
 }
 
+bool GoValidator::notAllPropsEmpty() {
+    return !exprProps_.inputProps().empty() || !exprProps_.varProps().empty() ||
+        !exprProps_.srcTagProps().empty() || !exprProps_.edgeProps().empty() ||
+        !exprProps_.dstTagProps().empty() || from_.fromType != FromType::kInstantExpr;
+}
+
```
CPWstatic, line 805:
```
!exprProps_.inputProps().empty() || !exprProps_.varProps().empty() ||
        !exprProps_.srcTagProps().empty() || !exprProps_.edgeProps().empty() ||
        !exprProps_.dstTagProps().empty()
```
Would it be better move this into ExprProps?
Then the logic in line 757 could be 
```
if (exprProps_.notAllPropsEmpty() || from_.fromType != FromType::kInstantExpr)
```

CPWstatic (approved):

yixinglu (approved):

PR #308, opened 2020-09-29
Extract filter node

jievince:
When the filter condition in the go statement contains not only SrcProperty but also DstProperty,  the GetNeighbors and Filter plan nodes will be far apart so that the FilterPushDownGetNbrRule cannot be applied to this situation.
This PR split the filter condition into two plan node that can be pushed down and those that cannot be pushed down so that FilterPushDownGetNbrRule can be applied to such a situation.

Maybe do it in Optimizer is more elegant.


yixinglu:
I don't think we need to do this optimization in validator. We should add some rules related to Filter to do same thing.

yixinglu:
https://github.com/vesoft-inc/nebula-graph/issues/339

PR #311, opened 2020-09-30
Add clone function for GetNeighbors/IndexScan plan node

yixinglu:
for usage in other opt rules.


src/planner/Query.cpp
```diff
@@ -27,6 +27,43 @@ std::unique_ptr<cpp2::PlanNodeDescription> Explore::explain() const {
...
 }
 
+GetNeighbors* GetNeighbors::clone(QueryContext* qctx) const {
+    auto newGN = GetNeighbors::make(qctx, nullptr, space_);
+    newGN->setSrc(qctx->objPool()->add(src_->clone().release()));
+    newGN->setEdgeTypes(edgeTypes_);
+    newGN->setEdgeDirection(edgeDirection_);
+    newGN->setDedup(dedup_);
+    newGN->setRandom(random_);
+    newGN->setLimit(limit_);
```
jievince, line 37: Why not set OrderBy?

yixinglu: miss it. Have fixed.

laura-ding (approved):

jievince (approved):

PR #312, opened 2020-09-30
Turn on the python test case for lookup

bright-starry-sky:


yixinglu (approved):

tests/query/stateless/test_lookup.py
```diff
@@ -18,12 +18,14 @@ def prepare(self):
...
 
     def test_edge_index(self):
+        time.sleep(self.delay)
         resp = self.execute('USE nbaLookup')
         self.check_resp_succeeded(resp)
         resp = self.execute("CREATE EDGE like(likeness int)")
         self.check_resp_succeeded(resp)
         resp = self.execute("CREATE EDGE serve(start_year int, end_year int)")
         self.check_resp_succeeded(resp)
+        time.sleep(self.delay)
```
laura-ding, line 28: why here need `sleep`

tests/query/stateless/test_lookup.py
```diff
@@ -33,97 +35,125 @@ def test_edge_index(self):
         resp = self.execute("CREATE EDGE INDEX like_index_1 on like(likeness)")
         self.check_resp_succeeded(resp)
         time.sleep(self.delay)
-        resp = self.execute('INSERT EDGE like(likeness) VALUES 100 -> 101:(95)')
+        resp = self.execute('INSERT EDGE like(likeness) VALUES "100" -> "101":(95)')
```
Shylock-Hg, line 38: In fact we will support the int vid, so I think better to copy the test file to `v2` folder and the modify.

tests/query/stateless/test_lookup.py
bright-starry-sky: Waiting meta client.

tests/query/stateless/test_lookup.py
bright-starry-sky: Good suggest. But I think it's better to put it in other PR once int VID is ready.

tests/query/stateless/test_lookup.py
bright-starry-sky: > why here need `sleep`

Need to get `Edge` from MetaClient when create edge index, So need to wait for data synchronization between `MetaClient` and `MetaServer`.

tests/query/stateless/test_lookup.py
Shylock-Hg: Until to now, we don't modify the origin test cases and add new in `stateless/v1`, so I think we'd better not to change it.

tests/query/stateless/test_lookup.py
laura-ding: I think the create index can't depend on the schema from the cache. it can check in metad, why check it by the cache.

bright-starry-sky: > I think the create index can't depend on the schema from the cache. it can check in metad, why check it by the cache.

You are right, This logic has been improved via #316 .I think it's best to add to regression test quickly.  Then fix the delay problem when the other PR is merged.

laura-ding (approved):

PR #316, opened 2020-10-09
support variable-length string type for index

bright-starry-sky:


bright-starry-sky:
Added override method visit for "CaseExpression" to compile.

src/optimizer/OptimizerUtils.cpp
```diff
@@ -437,5 +437,49 @@ Value OptimizerUtils::boundValueWithMin(const meta::cpp2::ColumnDef& col, const
...
+        case Value::Type::MAP:
+        case Value::Type::DATASET:
+        case Value::Type::PATH: {
+            DLOG(FATAL) << "Not supported value type " << type
+                        << "for index.";
+            return Value(NullType::BAD_TYPE);
+        }
+    }
+    DLOG(FATAL) << "Unknown value type " << static_cast<int>(type);
+    return Value(NullType::BAD_TYPE);
```
CPWstatic, line 481: Value::kNullBadType

src/validator/MaintainValidator.cpp
```diff
@@ -303,61 +302,46 @@ Status DropEdgeValidator::toPlan() {
...
-        if (schema_ == nullptr) {
-            return Status::SemanticError("No schema found for '%s'", tagName_.c_str());
-        }
-
-        status = IndexUtil::validateColumns(fields_);
-        if (!status.ok()) {
-            VLOG(1) << status;
-            break;
-        }
-    } while (false);
```
CPWstatic, line 327: Why remove these codes?

bright-starry-sky: > Why remove these codes?

No need to check in graph layer, this piece of code is already implemented in the meta layer.
And the meta cache's interval wait is avoided.

src/parser/MaintainSentences.cpp
```diff
@@ -238,9 +238,14 @@ std::string CreateTagIndexSentence::toString() const {
...
     buf += *tagName_;
     buf += " (";
-    std::string columns;
-    folly::join(", ", this->columns(), columns);
-    buf += columns;
+    std::string fields;
+    std::vector<std::string> fieldDefs;
+    for (const auto& field : this->fields()) {
+        std::string f = field.get_name();
+        if (field.__isset.type_length) f + "(" +field.get_type_length() + ")";
```
yixinglu, line 245: `fieldDefs.emplace_back(f+...)` ? 

Do you forget to use the string result?

src/parser/MaintainSentences.cpp
```diff
@@ -254,9 +259,14 @@ std::string CreateEdgeIndexSentence::toString() const {
...
     buf += *edgeName_;
     buf += " (";
-    std::string columns;
-    folly::join(", ", this->columns(), columns);
-    buf += columns;
+    std::string fields;
+    std::vector<std::string> fieldDefs;
+    for (const auto& field : this->fields()) {
+        std::string f = field.get_name();
+        if (field.__isset.type_length) f + "(" +field.get_type_length() + ")";
```
yixinglu, line 266: ditto

yixinglu (commented):
I have a question: Why do we need to specify the length of string column when creating index ?

src/parser/MaintainSentences.cpp
bright-starry-sky: > `fieldDefs.emplace_back(f+...)` ?
> 
> Do you forget to use the string result?

Good point. Thanks for reminding. fixed.

yixinglu (approved):

src/optimizer/OptimizerUtils.cpp
bright-starry-sky: > Value::kNullBadType

Good suggest. unified.

jievince (approved):
image

src/optimizer/OptimizerUtils.cpp
```diff
@@ -430,11 +430,55 @@ Value OptimizerUtils::boundValueWithMin(const meta::cpp2::ColumnDef& col, const
...
+            return Value::kNullBadType;
         }
     }
     DLOG(FATAL) << "Unknown value type " << static_cast<int>(type);
-    return Value(NullType::BAD_TYPE);
+    return Value::kNullBadType;
+}
+
+Value OptimizerUtils::normalizeValue(const meta::cpp2::ColumnDef& col, const Value& v) {
+    auto type = SchemaUtil::propTypeToValueType(col.get_type().get_type());
```
Shylock-Hg, line 441: Need check does `type` equal to `v.type()`?

src/parser/MaintainSentences.cpp
```diff
@@ -254,9 +260,15 @@ std::string CreateEdgeIndexSentence::toString() const {
     buf += " ON ";
     buf += *edgeName_;
     buf += " (";
-    std::string columns;
-    folly::join(", ", this->columns(), columns);
-    buf += columns;
+    std::vector<std::string> fieldDefs;
+    for (const auto& field : this->fields()) {
```
Shylock-Hg, line 264: I think you could move these code to `IndexFieldList::toString()`

src/optimizer/OptimizerUtils.cpp
bright-starry-sky: > Need check does `type` equal to `v.type()`?

Not need check at here, the type checked in another method . such as :
```
(root@nebula) [s]> lookup on tag1 where tag1.col1 == "row1_c1"
+------+
| _vid |
+------+
| "v1" |
+------+
Got 1 rows (time spent 1605/1987 us)

Wed, 04 Nov 2020 18:44:55 CST

(root@nebula) [cbs]> lookup on tag1 where tag1.col1 == 1
[ERROR (-12)]: SemanticError: expression error : tag1.col1

Wed, 04 Nov 2020 18:44:59 CST

(root@nebula) [s]> 
```

src/parser/parser.yy
```diff
@@ -1638,14 +1643,46 @@ drop_edge_sentence
...
+        delete $1;
+    }
+    ;
+
+index_field_list
+    : index_field {
+        $$ = new IndexFieldList();
+        std::unique_ptr<meta::cpp2::IndexFieldDef> field;
+        field.reset($1);
+        $$->addField(std::move(field));
```
Shylock-Hg, line 1668: I think you could accept raw pointer as parameter directly.

bright-starry-sky: > I think you could accept raw pointer as parameter directly.

I think here is ok. we don't need to keep this pointer. unique_ptr is enough.

src/parser/MaintainSentences.cpp
bright-starry-sky: > I think you could move these code to `IndexFieldList::toString()`

Yeah, I agree with your suggestion, this PR will solve the urgent compilation problem, which I think I will deal with later. thanks for your careful reading.

PR #322, opened 2020-10-10
add starts/ends-with expressions

Aiee:
1. add starts/ends-with expressions
2. add pytest for new expressions
3. fix an issue with pytest which causes tests to pass incorrectly (line numbers of actual&expect data should always be compared)

depend on https://github.com/vesoft-inc/nebula-storage/pull/159 & https://github.com/vesoft-inc/nebula-common/pull/253


CPWstatic (approved):

tests/query/v2/test_starts_ends_with.py
```diff
@@ -0,0 +1,300 @@
...
+# Copyright (c) 2020 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License,
+# attached with Common Clause Condition 1.0, found in the LICENSES directory.
+
+from tests.common.nebula_test_suite import NebulaTestSuite
+from tests.common.nebula_test_suite import T_NULL, T_EMPTY, T_NULL_BAD_TYPE
+import pytest
+
+class TestINandNotIn(NebulaTestSuite):
```
yixinglu, line 12: TestStartsWithAndEndsWith

tests/nebula-test-run.py
```diff
@@ -110,7 +110,7 @@ def run_tests(self, args):
         # Switch to your $src_dir/tests
         os.chdir(TEST_DIR)
         error_code = executor.run_tests(args)
-        data_loader.drop_data()
+        # data_loader.drop_data()
```
yixinglu, line 113: emm, maybe need to restore here.

yixinglu (approved):

darionyaphet:
It looks like a new grammar ?

Aiee:
> It looks like a new grammar ?

Correct.

.linters/cpp/checkKeyword.py
```diff
@@ -83,7 +83,10 @@
     'KW_BY',
     'KW_IN',
     'KW_NOT_IN',
+    'KW_CONTAINS',
```
laura-ding, line 86: It does not need to add here, here is reserved keywords

CPWstatic (approved):

yixinglu (approved):

PR #323, opened 2020-10-10
Fix the wrong results

laura-ding:


Aiee (dismissed):

src/executor/query/GetNeighborsExecutor.cpp
```diff
@@ -55,7 +55,7 @@ Status GetNeighborsExecutor::buildRequestDataSet() {
             continue;
         }
         auto ret = uniqueVid.emplace(val);
```
Shylock-Hg, line 57: Don't emplace to set if not require unique.

Shylock-Hg (approved):

yixinglu (approved):

PR #330, opened 2020-10-13
Fix edge test case failure with default prop value

yixinglu:
We should allow to insert edge with default prop values.


tests/maintain/test_tag_edge.py
```diff
@@ -548,7 +548,7 @@ def test_alter_edge_with_default(self):
 
         # insert without default prop, failed
         resp = self.client.execute('INSERT EDGE e() VALUES "1"->"2":()')
-        self.check_resp_failed(resp)
+        self.check_resp_succeeded(resp)
```
laura-ding, line 551: here without default value, it should be failed

PR #338, opened 2020-10-15
Remove the meaningless validator to report error.

Shylock-Hg:


CPWstatic (commented):
Why remove it?

Shylock-Hg:
> Why remove it?

Do we need it? When this happened fatal maybe better.

CPWstatic:
> > Why remove it?
> 
> Do we need it? When this happened fatal maybe better.

Prefer removing the DLOGs, and report error for unimplemented statements. For we have considered that we might not provide all implementation for the community.

Shylock-Hg:
> > > Why remove it?
> > 
> > 
> > Do we need it? When this happened fatal maybe better.
> 
> Prefer removing the DLOGs, and report error for unimplemented statements. For we have considered that we might not provide all implementation for the community.

Ok.

PR #344, opened 2020-10-19
Added planners.

CPWstatic:


src/planner/Planner.h
```diff
@@ -0,0 +1,45 @@
...
+public:
+    virtual ~Planner() = default;
+
+    static StatusOr<SubPlan> toPlan(Validator* validator);
+
+    virtual bool match(Validator* validator) = 0;
+
+    virtual StatusOr<SubPlan> transform(Validator* validator) = 0;
+
+    auto& plannersMap() {
```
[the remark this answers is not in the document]
CPWstatic: `Status::Error()` returns an instance of `Status`, it seems different with this scenario.

src/validator/SequentialValidator.h
```diff
@@ -8,12 +8,17 @@
...
 #include "common/base/Base.h"
-#include "validator/Validator.h"
 #include "parser/SequentialSentences.h"
+#include "validator/Validator.h"
 
 namespace nebula {
 namespace graph {
 
+struct SequentialAstContext final : AstContext {
+    QueryContext*                               qctx;
```
yixinglu, line 18: Maybe this qctx could be pulled up into AstContext.

src/planner/Planner.cpp
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#include "planner/Planner.h"
+
+#include "validator/Validator.h"
+
+namespace nebula {
+namespace graph {
+
+std::unordered_map<Sentence::Kind, std::vector<Planner*>> Planner::plannersMap_;
```
Shylock-Hg, line 14: Is it better to manage the Planners by tree or DAG?

yixinglu (approved):

src/planner/Planner.cpp
CPWstatic: Emm, the planners don't have that kind of relation.

yixinglu (commented):
LGTM

dutor (approved):
LGTM.

yixinglu (approved):

PR #347, opened 2020-10-20
Refine RewriteInputPropVisitor implementation

yixinglu:
Cleanup meaningless codes and replace fields movement in expression with setter.

Fix fixture usage and update the conftest.py position.


src/visitor/RewriteInputPropVisitor.cpp
```diff
@@ -121,42 +121,41 @@ void RewriteInputPropVisitor::visit(VariablePropertyExpression* expr) {
...
-    for (auto iter = items.begin(); iter < items.end(); ++iter) {
-        iter->second.get()->accept(this);
+    const auto& items = expr->items();
+    for (size_t i = 0; i < items.size(); ++i) {
+        auto key = items[i].first;
+        auto value = const_cast<Expression*>(items[i].second);
+        value->accept(this);
         if (ok()) {
-            *iter = std::make_pair(std::move(iter->first), std::move(result_));
+            expr->setItem(i, {std::make_unique<std::string>(*key), std::move(result_)});
```
CPWstatic, line 150: Maybe we could reuse `key`?
```
std::move(key)
```

src/visitor/RewriteInputPropVisitor.cpp
```diff
@@ -121,42 +121,41 @@ void RewriteInputPropVisitor::visit(VariablePropertyExpression* expr) {
...
 void RewriteInputPropVisitor::visit(SetExpression* expr) {
-    auto items = std::move(*expr).get();
-    for (auto iter = items.begin(); iter < items.end(); ++iter) {
-        iter->get()->accept(this);
+    const auto& items = expr->items();
+    for (size_t i = 0; i < items.size(); ++i) {
+        const_cast<Expression*>(items[i])->accept(this);
         if (ok()) {
-            *iter = std::move(result_);
+            expr->setItem(i, result_.release());
```
CPWstatic, line 138: Not sure if it would be helpful if we add such an interface
```
setItem(size_t index, std::unique_ptr<Expression> item)
```

yixinglu: Good suggestion. Fixed in vesoft-inc/nebula-common#268

src/visitor/RewriteInputPropVisitor.cpp
yixinglu: will fixed in #352

jievince (approved):
LGTM

nevermore3 (approved):

PR #349, opened 2020-10-21
Return semantic error in validator

yixinglu:
as title.


laura-ding (dismissed):

src/validator/GetSubgraphValidator.cpp
```diff
@@ -22,29 +22,19 @@ Status GetSubgraphValidator::validateImpl() {
     auto* gsSentence = static_cast<GetSubgraphSentence*>(sentence_);
 
     status = validateStep(gsSentence->step(), steps_);
```
Shylock-Hg, line 24: The `status` is useless.

yixinglu: fixed

src/validator/GetSubgraphValidator.cpp
```diff
@@ -21,30 +21,11 @@ Status GetSubgraphValidator::validateImpl() {
     Status status;
```
Shylock-Hg, line 21: Delete this define.

yixinglu: Good catch!

CPWstatic (approved):
Thanks.

jievince (approved):

PR #356, opened 2020-10-22
fix FoldConstantExprVisitor

jievince:


yixinglu (commented):
Thanks, Need some UTs.

PR #365, opened 2020-10-26
Support text search

bright-starry-sky:
Depends on:
 https://github.com/vesoft-inc/nebula-common/pull/260

### DIFF
--- a/src/executor/admin/SpaceExecutor.cpp
+++ b/src/executor/admin/SpaceExecutor.cpp
@@ -10,9 +10,6 @@
 #include "planner/Admin.h"
 #include "util/SchemaUtil.h"
 #include "util/ScopedTimer.h"
-#include "common/plugin/fulltext/elasticsearch/ESGraphAdapter.h"
-
-DECLARE_uint32(ft_request_retry_times);
 
 namespace nebula {
 namespace graph {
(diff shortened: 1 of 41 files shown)


src/executor/admin/SpaceExecutor.cpp
```diff
@@ -10,24 +10,62 @@
...
+                        return Status::Error("external index create failed : %s",
+                                             edgeTSIndex.c_str());
+                    }
+
+                    auto tagTSIndex = nebula::plugin::IndexTraits::indexName(spaceName, false);
+                    if (createTSIndex(ret.value(), tagTSIndex)) {
+                        return Status::Error("external index create failed : %s",
+                                             tagTSIndex.c_str());
+                    }
+                }
```
critical27, line 46: I am not pretty sure we should create index here. Why we don't `CREATE FULLTEXT TAG | EDGE INDEX index_name ON space_name [fulltext_clause]` like in design doc?

src/parser/parser.yy
```diff
@@ -1048,8 +1073,245 @@ match_return
...
+    }
+    ;
+
+sign_out_text_search_service_sentence
+    : KW_SIGN KW_OUT KW_TEXT KW_SERVICE {
+        $$ = new SignOutTextServiceSentence();
+    }
+    ;
+
+text_search_argument
```
critical27, line 1120: We could add some doc later of how to use full text index

src/validator/IndexScanValidator.cpp
```diff
@@ -83,15 +97,148 @@ Status IndexScanValidator::prepareYield() {
...
+                if (expr->arg()->fuzziness() == -1) {
+                    fuzz = "AUTO";
+                } else {
+                    fuzz = expr->arg()->fuzziness();
+                }
+                suc = nebula::plugin::ESGraphAdapter::kAdapter->fuzzy(ftc,
+                                                                      doc,
+                                                                      limit,
+                                                                      fuzz,
+                                                                      *expr->arg()->op(),
```
critical27, line 203: What is op, and/or?

src/executor/admin/SpaceExecutor.cpp
bright-starry-sky: > I am not pretty sure we should create index here. Why we don't `CREATE FULLTEXT TAG | EDGE INDEX index_name ON space_name [fulltext_clause]` like in design doc?

Good question. the original design is similar to what you said.after an offline discussion with Sherman. I modified it to create external indexes and paces together. For two reasons: 1, external indexes are associated with space, not with tag or edge; 2, external index creation does not require meta; the graph layer itself is ok.

src/validator/IndexScanValidator.cpp
bright-starry-sky: > What is op, and/or?

Yes, same with match api of ElasticSearch.

bright-starry-sky:
Done. ready to review.

src/validator/IndexScanValidator.cpp
```diff
@@ -93,26 +95,173 @@ Status IndexScanValidator::prepareFilter() {
...
+        return relExpr;
+    } else {
+        return genOrFilterFromList(tsExpr, vRet.value());
+    }
+    return Status::OK();
+}
+
+std::unique_ptr<Expression> IndexScanValidator::genOrFilterFromList(
+    TextSearchExpression* expr, const std::vector<std::string>& values) {
+    auto filter = std::make_unique<LogicalExpression>(
```
critical27, line 151: It seems `LogicalExpression` has support multiple operands, could we use it?

critical27: BTW, have test the performance between pass as filter or pass by  multi`IndexQueryContext`

PR #374, opened 2020-10-28
Fix cluster action

yixinglu:
Add backoffLimit for job to limit job restarts
Sleep some time for waiting graph service ready

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -136,6 +136,6 @@ jobs:
       - name: Upload logs
         uses: actions/upload-artifact@v2
         if: ${{ failure() }}
-        with:	
+        with:
           name: ${{ matrix.os }}-${{ matrix.compiler }}-nebula-test-logs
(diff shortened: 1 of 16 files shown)


laura-ding (dismissed):

src/service/QueryInstance.cpp
```diff
@@ -133,22 +133,7 @@ void QueryInstance::onError(Status status) {
         case Status::Code::kPermissionError:
             rctx->resp().set_error_code(cpp2::ErrorCode::E_BAD_PERMISSION);
             break;
-        case Status::Code::kBalanced:
```
Shylock-Hg, line 136: Why change this ?

yixinglu: We need not to update here each time when storage add new error status. Only handling our concerned error code is enough.

dutor:
Please resolve the conflicts

nevermore3 (approved):

jievince (approved):

PR #379, opened 2020-10-28
Fold the function call itself if it's stateless.

Shylock-Hg:
Follow https://github.com/vesoft-inc/nebula-common/pull/281


src/visitor/FoldConstantExprVisitor.cpp
```diff
@@ -75,6 +77,14 @@ void FoldConstantExprVisitor::visit(FunctionCallExpression *expr) {
             }
         }
     }
+    auto result = FunctionManager::getStateless(*expr->name(), expr->args()->args().size());
```
yixinglu, line 80: prefer `sideEffect` or `isPure` than `stateless` when we talk about function

dutor: > prefer `sideEffect` or `isPure` than `stateless` when we talk about function

Agreed. `isPure` is better.

Shylock-Hg: Ok, fixed.

dutor (approved):
Great fix!

yixinglu (approved):

PR #380, opened 2020-10-29
Support IN/BOTH direction in MATCH

dutor:
As title.


src/validator/MatchValidator.cpp
```diff
@@ -482,6 +482,9 @@ Status MatchValidator::buildStep() {
     if (!edgeInfo.edgeTypes.empty()) {
         for (auto edgeType : edgeInfo.edgeTypes) {
             EdgeProp edgeProp;
+            if (edgeInfo.direction == Direction::IN_EDGE) {
```
laura-ding, line 485: the EdgeDirection::BOTH?

dutor: > the EdgeDirection::BOTH?

Will be set below.

dutor: addressed

laura-ding (approved):

yixinglu (approved):

PR #385, opened 2020-10-31
Add DISTINCT & ORDER BY & SKIP & LIMIT

dutor:
TODO: `Sort` needs to be augmented to support general expressions as order factors.

### DIFF
--- a/src/planner/planners/MatchSolver.cpp
+++ b/src/planner/planners/MatchSolver.cpp
@@ -14,6 +14,8 @@ namespace graph {
 Status MatchSolver::buildReturn(MatchAstContext* matchCtx, SubPlan& subPlan) {
     auto *yields = new YieldColumns();
     std::vector<std::string> colNames;
+    auto *sentence = static_cast<MatchSentence*>(matchCtx->sentence);
+    PlanNode *current = subPlan.root;
 
     for (auto *col : matchCtx->yieldColumns->columns()) {
         auto kind = col->expr()->kind();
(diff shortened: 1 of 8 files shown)


CPWstatic (approved):

jievince (approved):

src/planner/planners/MatchSolver.cpp
```diff
@@ -45,10 +47,68 @@ Status MatchSolver::buildReturn(MatchAstContext* matchCtx, SubPlan& subPlan) {
...
+            if (factor->expr()->kind() != Expression::Kind::kLabel) {
+                return Status::SemanticError("Only column name can be used as sort item");
+            }
+            auto *name = static_cast<const LabelExpression*>(factor->expr())->name();
+            auto iter = inputColIndices.find(*name);
+            if (iter == inputColIndices.end()) {
+                return Status::SemanticError("Column `%s' not found", name->c_str());
+            }
+            indexedFactors.emplace_back(iter->second, factor->orderType());
+        }
```
CPWstatic, line 84: These could be part of semantic analysis.

src/planner/planners/MatchSolver.cpp
```diff
@@ -45,10 +47,68 @@ Status MatchSolver::buildReturn(MatchAstContext* matchCtx, SubPlan& subPlan) {
...
+        int64_t offset = 0;
+        int64_t count = std::numeric_limits<int64_t>::max();
+        if (skipExpr != nullptr) {
+            auto *constant = static_cast<const ConstantExpression*>(skipExpr);
+            offset = constant->value().getInt();
+        }
+        if (limitExpr != nullptr) {
+            auto *constant = static_cast<const ConstantExpression*>(limitExpr);
+            count = constant->value().getInt();
+        }
```
CPWstatic, line 104: ditto.

src/planner/planners/MatchSolver.cpp
dutor: Yes. But then I have to store some states somewhere so that I can access them later. So far the only place is the MatchValidator instance. I think that will make the later refactoring more tedious. We can extract them out when we add a separate context for each clause.

CPWstatic: We could use `MatchAstContext` for now.

yixinglu (dismissed):
LGTM

src/validator/MatchValidator.cpp
```diff
@@ -207,6 +194,66 @@ Status MatchValidator::validateReturn(MatchReturn *ret) {
     }
     NG_RETURN_IF_ERROR(validateAliases(exprs));
 
+    auto *skipExpr = ret->skip();
+    auto *limitExpr = ret->limit();
+    int64_t skip = 0;
+    int64_t limit = std::numeric_limits<int64_t>::max();
+    if (skipExpr != nullptr) {
+        if (skipExpr->kind() != Expression::Kind::kConstant) {
```
laura-ding, line 202: it can't be a mathematical expression？maybe you can use `evaluableExpr` to check it.

src/validator/MatchValidator.cpp
```diff
@@ -207,6 +194,66 @@ Status MatchValidator::validateReturn(MatchReturn *ret) {
...
+        }
+        auto *constant = static_cast<const ConstantExpression*>(skipExpr);
+        if (!constant->value().isInt()) {
+            return Status::SemanticError("SKIP should be an integral constant");
+        }
+        skip = constant->value().getInt();
+    }
+
+    if (limitExpr != nullptr) {
+        if (limitExpr->kind() != Expression::Kind::kConstant) {
```
laura-ding, line 213: ditto

src/validator/MatchValidator.cpp
dutor: Maybe next time.

dutor: done.

CPWstatic (approved):

yixinglu (approved):

PR #392, opened 2020-11-03
Handle path build expr.

CPWstatic:


nevermore3 (approved):

nevermore3 (dismissed):

src/visitor/FoldConstantExprVisitor.cpp
```diff
@@ -296,5 +296,21 @@ Expression *FoldConstantExprVisitor::fold(Expression *expr) const {
...
+    for (size_t i = 0; i < items.size(); ++i) {
+        auto item = items[i].get();
+        item->accept(this);
+        if (!canBeFolded_) {
+            canBeFolded = false;
+            continue;
+        }
+        if (item->kind() != Expression::Kind::kConstant) {
+            expr->setItem(i, std::unique_ptr<Expression>{fold(item)});
+        }
```
jievince, line 311: It's better to use isConstant() to check if item is a constant expression. If not, then item->accept(this)

CPWstatic: ok.

src/visitor/ExprVisitorImpl.cpp
```diff
@@ -125,5 +125,14 @@ void ExprVisitorImpl::visitBinaryExpr(BinaryExpression *expr) {
     }
 }
 
+void ExprVisitorImpl::visit(PathBuildExpression *expr) {
+    DCHECK(ok());
```
laura-ding, line 129: it must true?

jievince (approved):
image

src/visitor/ExprVisitorImpl.cpp
CPWstatic: Yes, the father expr should not call `accept` once it is not ok.

nevermore3 (approved):

PR #398, opened 2020-11-06
support  index without fields

panda-sheep:
As title


src/parser/test/ParserTest.cpp
```diff
@@ -474,6 +474,12 @@ TEST(Parser, ColumnSpacesTest) {
 }
 
 TEST(Parser, IndexOperation) {
+    {
+        GQLParser parser;
+        std::string query = "CREATE TAG INDEX empty_field_index ON person()";
```
Shylock-Hg, line 479: What's the meaning for empty field index?

panda-sheep: refer to https://github.com/vesoft-inc/nebula-storage/pull/174

use to get all vertices on tag or edges on one edgetype

Shylock-Hg: Emm.Mix the sentence to create properties index and tag/edge index is a little ambiguous

src/parser/parser.yy
```diff
@@ -1736,9 +1739,11 @@ index_field
 index_field_list
     : index_field {
         $$ = new IndexFieldList();
-        std::unique_ptr<meta::cpp2::IndexFieldDef> field;
-        field.reset($1);
-        $$->addField(std::move(field));
+        if ($1) {
```
Shylock-Hg, line 1742: I think you could remove the if statement refer to the rule like `opt_argument_list`

src/parser/test/ParserTest.cpp
panda-sheep: I think it's ok.
You think it just creating an index of 0 fields to get the vertices of the tag or the edges on the edgeType.

Similar to the logic of a normal index, when an index with one or more fields is included, it is not allowed to create an index with zero fields.

src/parser/parser.yy
panda-sheep: Good, fixed!

yixinglu (commented):
This PR is good to me. But I only want to know which field the index created on when user don't specify the filed of tag or edge.

panda-sheep:
> This PR is good to me. But I only want to know which field the index created on when user don't specify the filed of tag or edge.

Good question.
We could think of it as a special index, which is only used to get the ` `of the tag or the `src`, `rank`, `dst `of the edge

yixinglu (approved):

laura-ding (approved):

PR #408, opened 2020-11-10
Support Unwind & With

jievince:
Currently, support any combination of `with` clause and `unwind` clause. 
Any combination with `match` clause will be supported later
https://neo4j.com/docs/cypher-manual/current/clauses/unwind/


src/executor/query/UnwindExecutor.cpp
```diff
@@ -0,0 +1,65 @@
...
+    auto *unwind = asNode<Unwind>(node());
+    auto *expr = const_cast<Expression *>(unwind->expr());
+
+    auto iter = ectx_->getResult(unwind->inputVar()).iter();
+    DCHECK(!!iter);
+    QueryExpressionContext ctx(ectx_);
+
+    DataSet ds;
+    ds.colNames = unwind->colNames();
+    if (expr->kind() == Expression::Kind::kVarProperty) {   // UNWIND var AS r
```
yixinglu, line 28: Are you sure the expression kind supported only contains `VarProperty` and `ConstantExpression` ? How do you handle the scenario like `var[0]`.

src/executor/query/UnwindExecutor.cpp
```diff
@@ -0,0 +1,65 @@
...
+                ds.rows.emplace_back(Row({std::move(v)}));
+            }
+        }
+    } else {   // UNWIND [1, 2, 3] AS r
+        Value val = expr->eval(ctx);
+        std::vector<Value> vals = extractList(val);
+        for (const auto &v : vals) {
+            ds.rows.emplace_back(Row({std::move(v)}));
+        }
+    }
```
CPWstatic, line 43: Please investigate carefully of the `unwind` behave.
```
match (a)-[e*1..2]-() unwind [1,2,3] as c return c
```
```
unwind [1,2,3] as c return c
```
```
unwind [1,2,3] as c match (a{name:c})-[e*1..2]-() return e,c
```

src/planner/planners/MatchSolver.cpp
```diff
@@ -81,6 +81,26 @@ Status MatchSolver::buildReturn(MatchAstContext* mctx, SubPlan& subPlan) {
     return Status::OK();
 }
 
+Status MatchSolver::buildUnwind(MatchAstContext *mctx, UnwindClause &clause, SubPlan &subPlan) {
```
CPWstatic, line 84: Where do you want to use it?

src/executor/query/UnwindExecutor.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+std::vector<Value> UnwindExecutor::extractList(Value &val) {
+    std::vector<Value> ret;
+    if (val.isList()) {
+        auto &list = val.getList();
+        for (size_t i = 0; i < list.size(); ++i) {
+            ret.emplace_back(std::move(list[i]));
+        }
+    } else {
+        if (val != Value::kNullValue) {
```
laura-ding, line 50: Maybe is `if (!(val.isNull() || val.isEmpty()))`

src/planner/match/UnwindClausePlanner.cpp
```diff
@@ -0,0 +1,66 @@
...
+        yields->addColumn(newColumn);
+        if (col->alias() != nullptr) {
+            colNames.emplace_back(*col->alias());
+        } else {
+            colNames.emplace_back(col->expr()->toString());
+        }
+    }
+
+    auto* unwind = Unwind::make(uctx->qctx, nullptr, yields);
+    unwind->setColNames(colNames);
```
laura-ding, line 58: unwind->setColNames(std::move(colNames));

src/validator/MatchValidator.cpp
```diff
@@ -204,14 +246,19 @@ Status MatchValidator::validateFilter(const Expression *filter,
...
 
     return Status::OK();
 }
 
 Status MatchValidator::validateReturn(MatchReturn *ret,
-                                      const MatchClauseContext &matchClauseCtx,
+                                      const CypherClauseContextBase *cypherClauseCtx,
                                       ReturnClauseContext &retClauseCtx) const {
+    auto kind = cypherClauseCtx->kind;
+    if (kind != CypherClauseKind::kMatch && kind != CypherClauseKind::kUnwind &&
```
laura-ding, line 258: why not use `switch case` replace. then just need to maintain one place.

src/validator/MatchValidator.cpp
```diff
@@ -382,6 +443,117 @@ Status MatchValidator::validateStepRange(const MatchStepRange *range) const {
...
+        }
+        if (value.getInt() < 0) {
+            return Status::SemanticError("SKIP should not be negative");
+        }
+        skip = value.getInt();
+    }
+
+    if (limitExpr != nullptr) {
+        if (!evaluableExpr(limitExpr)) {
+            return Status::SemanticError("SKIP should be instantly evaluable");
```
laura-ding, line 486: Limit

src/executor/query/UnwindExecutor.cpp
jievince: You're right. Thanks.

src/validator/MatchValidator.cpp
jievince: Thanks, will be fixed.

src/context/ast/QueryAstContext.h
```diff
@@ -96,17 +96,20 @@ struct ReturnClauseContext final : CypherClauseContextBase {
...
 struct WithClauseContext final : CypherClauseContextBase {
     WithClauseContext() : CypherClauseContextBase(CypherClauseKind::kWith) {}
 
+    bool                                        distinct{false};
     const YieldColumns*                         yieldColumns{nullptr};
     std::unique_ptr<OrderByClauseContext>       order;
     std::unique_ptr<PaginationContext>          pagination;
     std::unique_ptr<WhereClauseContext>         where;
+    std::unordered_map<std::string, AliasType>* aliasesPtr{nullptr};
+    std::unordered_map<std::string, AliasType>  aliases;
```
CPWstatic, line 112: Why we need both value and ptr?

src/planner/match/MatchSolver.h
```diff
@@ -30,7 +30,7 @@ class MatchSolver final {
 
     static Expression* rewrite(const LabelAttributeExpression* la);
 
-    static Expression* doRewrite(const std::unordered_map<std::string, AliasType>& aliases,
+    static Expression* doRewrite(const std::unordered_map<std::string, AliasType>* aliases,
```
CPWstatic, line 33: Why replace it with ptr? i don't see any improvement.

src/validator/MatchValidator.cpp
```diff
@@ -23,40 +23,89 @@ AstContext* MatchValidator::getAstContext() {
...
     auto &clauses = sentence->clauses();
 
-    if (clauses.size() > 1) {
-        return Status::SemanticError("Multi clause MATCH not supported");
-    }
+    std::unordered_map<std::string, AliasType> *curAliasesPtr = nullptr;
+    YieldColumns *curYieldColumns = nullptr;
+    auto retClauseCtx = getContext<ReturnClauseContext>();
+    for (size_t i = 0; i < clauses.size(); ++i) {
+        if (clauses[i]->isMatch()) {
```
CPWstatic, line 34: Switch

jievince: OK

src/context/ast/QueryAstContext.h
jievince: aliasesPtr points to the previous clause's aliases. It's used when checking if a label used in the current clause is legal.

src/validator/MatchValidator.cpp
jievince: ok

tests/tck/features/match/WithUnwind.feature
```diff
@@ -0,0 +1,140 @@
...
+      UNWIND [1, 2, 3] AS a
+      WITH "hello" AS b, a AS c
+      UNWIND c as d
+      RETURN b, c, d
+      """
+    Then the result should be, in any order, with relax comparison:
+      | b       | c | d |
+      | "hello" | 1 | 1 |
+      | "hello" | 2 | 2 |
+      | "hello" | 3 | 3 |
```
yixinglu, line 140: why not result is:
```text
      | b       | c       | d |
      | "hello" | [1,2,3] | 1 |
      | "hello" | [1,2,3] | 2 |
      | "hello" | [1,2,3] | 3 |
```

src/context/ast/QueryAstContext.h
yixinglu: If so, please given a meaningful name

jievince: Maybe we can call it `previousAliases`

src/executor/query/UnwindExecutor.cpp
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    return finish(ResultBuilder().value(Value(std::move(ds))).finish());
+}
+
+std::vector<Value> UnwindExecutor::extractList(Value &val) {
+    std::vector<Value> ret;
+    if (val.isList()) {
+        auto &list = val.getList();
```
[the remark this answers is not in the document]
jievince: ok

tests/tck/features/match/WithUnwind.feature
jievince: Then original query is:
```
  UNWIND [1, 2, 3] AS a
  WITH "hello" AS b, a AS c
  UNWIND c as d
  RETURN b, c, d
```
The output of the `with` clause is:
```
╒═══════╤═══╕
│"b"    │"c"│
╞═══════╪═══╡
│"hello"│1  │
├───────┼───┤
│"hello"│2  │
├───────┼───┤
│"hello"│3  │
└───────┴───┘
```
So for the second `unwind` clause, label `c` is a dataset that has 3 rows, and each row has an integer(No `list` here!).
The cypher doc says:
```
7. Using UNWIND with an expression that is not a list
Using UNWIND on an expression that does not return a list, will return the same result as using UNWIND on a list that just contains that expression. As an example, UNWIND 5 is effectively equivalent to UNWIND[5]. The exception to this is when the expression returns null — this will reduce the number of rows to zero, causing it to cease its execution and return no results.
```

src/context/ast/QueryAstContext.h
jievince: > Why we need both value and ptr?

`value` means aliases that the current clause produces, and `ptr` means aliases that the current clause may use(it's produced by last clause)

CPWstatic: > Maybe we can call it `previousAliases`

aliasesUsed, aliasesGenerated

src/planner/match/MatchSolver.h
jievince: eg. wctx->aliases might be a nullptr, so I need add DCHECK(wctx->aliases) before call doRewrite, and  everytime I call doRewrite I have to add this DCHECK. 
So I replace this argument with ptr, and so I can add DCHECK in the function body of doRewrite
```
    auto rewriter = [wctx](const Expression* expr) {
        return MatchSolver::doRewrite(*wctx->aliases, expr);
    };
```

CPWstatic (commented):
Excellent work!

src/validator/MatchValidator.cpp
```diff
@@ -336,8 +421,9 @@ Status MatchValidator::validateReturn(MatchReturn *ret,
     return Status::OK();
 }
 
-Status MatchValidator::validateAliases(const std::vector<const Expression *> &exprs,
-                                       std::unordered_map<std::string, AliasType> &aliases) const {
+Status MatchValidator::validateAliases(
+    const std::vector<const Expression *> &exprs,
+    std::unordered_map<std::string, AliasType> *aliasesUsed) const {
```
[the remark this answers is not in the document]
jievince: yes, const is better. I'll fix it

src/executor/test/UnwindTest.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "context/QueryContext.h"
+#include "executor/query/UnwindExecutor.h"
+#include "executor/test/QueryTestBase.h"
+#include "planner/Logic.h"
+#include "planner/Query.h"
+
+namespace nebula {
+namespace graph {
+
+class UnwindTest : public QueryTestBase {
```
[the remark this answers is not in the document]
jievince: OK

src/validator/MatchValidator.cpp
```diff
@@ -358,7 +444,7 @@ Status MatchValidator::validateAliases(const std::vector<const Expression *> &ex
                 name = static_cast<const LabelAttributeExpression*>(refExpr)->left()->name();
             }
             DCHECK(name != nullptr);
-            if (aliases.count(*name) != 1) {
+            if (!aliasesUsed || aliasesUsed->count(*name) != 1) {
```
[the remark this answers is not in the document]
jievince: When `aliasesUsed` is nullptr, it can validate such qeuries:
```
Unwind a as b return b;    // query 1
Unwind [1,2,3] as c return c; // query 2
```
The `aliasesUsed` for these two queries are nullptr, but `query 1` uses a label `a`,  `query 2` uses a constant `[1,2,3]`, So `validateAliases` can know that  `query 1` is illegal, and `query 2` is legal.

src/validator/MatchValidator.cpp
```diff
@@ -382,6 +468,120 @@ Status MatchValidator::validateStepRange(const MatchStepRange *range) const {
...
+            }
+            auto *name = static_cast<const LabelExpression *>(factor->expr())->name();
+            auto iter = inputColIndices.find(*name);
+            if (iter == inputColIndices.end()) {
+                return Status::SemanticError("Column `%s' not found", name->c_str());
+            }
+            withClauseCtx.order->indexedOrderFactors.emplace_back(iter->second,
+                                                                  factor->orderType());
+        }
+    }
```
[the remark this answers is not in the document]
jievince: can use a Macro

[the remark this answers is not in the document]
CPWstatic: Emm, macro is not a good solution, make them function like:
```
validatePagination
validateOrderBy
```

src/validator/MatchValidator.cpp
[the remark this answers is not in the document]
CPWstatic: Got it.

CPWstatic (approved):
LGTM.

tests/tck/features/match/WithUnwind.feature
yixinglu: very clear, Got it.

src/validator/MatchValidator.cpp
```diff
@@ -23,47 +23,113 @@ AstContext* MatchValidator::getAstContext() {
...
     auto &clauses = sentence->clauses();
 
-    if (clauses.size() > 1) {
-        return Status::SemanticError("Multi clause MATCH not supported");
-    }
+    std::unordered_map<std::string, AliasType> *aliasesUsed = nullptr;
+    YieldColumns *prevYieldColumns = nullptr;
+    auto retClauseCtx = getContext<ReturnClauseContext>();
+    for (size_t i = 0; i < clauses.size(); ++i) {
+        switch (clauses[i]->kind()) {
```
yixinglu, line 34: Why not extract some common functions for the huge switch check?

src/validator/MatchValidator.cpp
```diff
@@ -23,47 +23,113 @@ AstContext* MatchValidator::getAstContext() {
...
+                }
+
+                aliasesUsed = &withClauseCtx->aliasesGenerated;
+                prevYieldColumns = const_cast<YieldColumns *>(withClauseCtx->yieldColumns);
+
+                if (i == clauses.size() - 1) {
+                    retClauseCtx->aliasesUsed = aliasesUsed;
+                    NG_RETURN_IF_ERROR(
+                        validateReturn(sentence->ret(), withClauseCtx.get(), *retClauseCtx));
+                }
```
yixinglu, line 115: If all cases have to vailidate return sentence, why not pull this check outside the switch block?

tests/tck/features/match/WithUnwind.feature
```diff
@@ -0,0 +1,140 @@
...
+    When executing query:
+      """
+      WITH [1, 2, 3] AS a WITH a AS a, "hello" AS b
+      RETURN a, b
+      """
+    Then the result should be, in any order, with relax comparison:
+      | a       | b       |
+      | [1,2,3] | "hello" |
+
+  @skip
```
yixinglu, line 24: We haven't support with and match combination now ?

src/context/ast/QueryAstContext.h
```diff
@@ -115,14 +118,18 @@ struct MatchClauseContext final : CypherClauseContextBase {
 
     std::vector<NodeInfo>                       nodeInfos;
     std::vector<EdgeInfo>                       edgeInfos;
-    std::unordered_map<std::string, AliasType>  aliases;
     std::unique_ptr<PathBuildExpression>        pathBuild;
     std::unique_ptr<WhereClauseContext>         where;
+    std::unordered_map<std::string, AliasType>* aliasesUsed{nullptr};
```
yixinglu, line 123: Could we pull the field up to the base context ?

src/validator/MatchValidator.cpp
jievince:
```
NG_RETURN_IF_ERROR(
                        validateReturn(sentence->ret(), withClauseCtx.get(), *retClauseCtx));
matchCtx_->clauses.emplace_back(std::move(withClauseCtx));
```
Because return clause uses the preceding clause to call `validateReturn`, so if pull validateREturn outside the switch block, I can't std::move(the_preceding_clause).

src/validator/MatchValidator.cpp
jievince: For common codes, I have extracted them into functions. Such as `combineAliases`.  For the other codes, I coudn't extract them.

tests/tck/features/match/WithUnwind.feature
jievince: Yes, they will be in another PR.

src/planner/match/SegmentsConnector.cpp
```diff
@@ -16,9 +17,16 @@ StatusOr<SubPlan> SegmentsConnector::connectSegments(CypherClauseContextBase* le
                                                      CypherClauseContextBase* rightCtx,
                                                      SubPlan& left,
                                                      SubPlan& right) {
-    if (leftCtx->kind == CypherClauseKind::kReturn && rightCtx->kind == CypherClauseKind::kMatch) {
+    UNUSED(rightCtx);
```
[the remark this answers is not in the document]
jievince: I'll fix it.

src/context/ast/QueryAstContext.h
jievince: Currently, OrderByClauseContext and PaginationContext couldn't refer to variables produced by the preceding clauses.
I would like to handle this situation in the PR of 'unwind match'

yixinglu (approved):

PR #409, opened 2020-11-10
support lookup on tagName/edgeName

panda-sheep:
The following syntax is supported:
```
lookup on tagName
lookup on edgeName
```
Where and yield clauses are not allowed at this time.


src/optimizer/rule/IndexScanRule.cpp
```diff
@@ -402,6 +430,16 @@ IndexItem IndexScanRule::findOptimalIndex(graph::QueryContext *qctx,
...
+// Only use "lookup on tagname"
+IndexItem IndexScanRule::findOptimalIndex(graph::QueryContext *qctx,
+                                          const OptGroupNode *groupNode) const {
+    auto indexes = allIndexesBySchema(qctx, groupNode);
+    if (indexes.empty()) {
+        return nullptr;
+    }
+    return indexes[0];
+}
+
```
bright-starry-sky, line 442: A suggestion . rename the method to `findLightestIndex`, and find the lightest index. for example : 
index1 on tag ()
index2 on tag(c1)
index3 on tag(c1,c2,...)
Index1 should be selected. right?

bright-starry-sky:
Except one suggestion of above, totally is good for me.

bright-starry-sky:
Emm, I think you need to check the yield clause. the yield clause should be empty when the filter is empty. right?

src/optimizer/rule/IndexScanRule.cpp
panda-sheep: The current strategy is to find an index randomly.

panda-sheep:
> Emm, I think you need to check the yield clause. the yield clause should be empty when the filter is empty. right?

Good question， IndexScanValidator::prepareYield has already handled.

bright-starry-sky:
> > Emm, I think you need to check the yield clause. the yield clause should be empty when the filter is empty. right?
> 
> Good question， IndexScanValidator::prepareYield has already handled.

Got. I saw. my mistake.

src/optimizer/rule/IndexScanRule.cpp
panda-sheep: Here, the first index is used as the index to be used for returning.

bright-starry-sky: > Here, the first index is used as the index to be used for returning.

I think, the best scanning performance of the storage layer is the lightest index. WDYT?
In addition, if return the first index, this method name is best `getFirstIndex` or `findAnyIndex` . right?

panda-sheep: Make sense,  Let’s use the index with the fewest fields, 👍

panda-sheep: Fixed.

dutor (commented):
Thanks!

I have some questions about the design:
 1. If the filter is not present, surely we could scan by label. But what if the filter is provided while no applicable index can be used?
 2. Why can't we return properties while scanning by label?

bright-starry-sky (approved):
LGTM.

Shylock-Hg:
How to scan label with properties filter?

panda-sheep:
> Thanks!
> 
> I have some questions about the design:
> 
> 1. If the filter is not present, surely we could scan by label. But what if the filter is provided while no applicable index can be used?
> 2. Why can't we return properties while scanning by label?


`lookup on tagName  `      Use index to return tag's  all  `vid `by default
lookup on edgeName       Use index to return the `src, rank, dst `   of the all  edges of one edgetype  by default

panda-sheep:
> > Thanks!
> > I have some questions about the design:
> > 
> > 1. If the filter is not present, surely we could scan by label. But what if the filter is provided while no applicable index can be used?
> > 2. Why can't we return properties while scanning by label?
> 
> `lookup on tagName ` Use index to return tag's all `vid `by default
> lookup on edgeName Use index to return the `src, rank, dst ` of the all edges of one edgetype by default
> 
> So no where clause is needed, No need for yield clause

dutor:
> > Thanks!
> > I have some questions about the design:
> > 
> > 1. If the filter is not present, surely we could scan by label. But what if the filter is provided while no applicable index can be used?
> > 2. Why can't we return properties while scanning by label?
> 
> `lookup on tagName ` Use index to return tag's all `vid `by default
> `lookup on edgeName` Use index to return the `src, rank, dst ` of the all edges of one edgetype by default
> 
> So no` where clause` is needed, No need for `yield clause`

You didn't answer any of my questions.

panda-sheep:
> > > Thanks!
> > > I have some questions about the design:
> > > 
> > > 1. If the filter is not present, surely we could scan by label. But what if the filter is provided while no applicable index can be used?

> > > 2. Why can't we return properties while scanning by label?
> > 
ok
1） Before this pr， if the current `lookup `statement does not have a `where `condition, an error will be reported
If a filter is provided without an applicable index, it will report that there is no valid index.

2)  When there is no valid index, we do not scan the data.  I think you may confuse the where statement with the filter.

dutor (approved):

PR #412, opened 2020-11-11
Fix test dir not exist when run single test

laura-ding:


tests/common/nebula_service.py
```diff
@@ -22,7 +22,7 @@ class NebulaService(object):
     def __init__(self, build_dir, src_dir):
         self.build_dir = build_dir
         self.src_dir = src_dir
-        self.work_dir = os.path.join(os.getcwd(), 'build', 'server')
+        self.work_dir = os.path.join(os.getcwd(), 'build', 'server_' + time.strftime("%Y-%m-%dT%H:%M:%S", time.localtime()) )
```
jievince, line 25: Why not use `self.work_dir = os.path.join(build_dir, 'server_.....')`?  If so, `os.mkdir` will work well.

tests/common/nebula_service.py
```diff
@@ -85,7 +85,7 @@ def _telnet_port(self, port):
     def install(self):
         if os.path.exists(self.work_dir):
             shutil.rmtree(self.work_dir)
-        os.mkdir(self.work_dir)
+        os.makedirs(self.work_dir)
```
Shylock-Hg, line 88: Don't do this, it's feature that don't allow to create server if `build` folder don't exists.

tests/common/nebula_service.py
laura-ding: I understand that was running the test, where generating the dir. Your understanding is to be generated in a fixed place. it ok for them.

tests/common/nebula_service.py
laura-ding: So your path generation is wrong, you created it with the current directory, not the build directory as the parent node. you can make it as @jievince said.

Shylock-Hg: Yes, build_dir better.

jievince (approved):

dutor (approved):

PR #413, opened 2020-11-11
forbid range scan of bool in lookup

critical27:
Some prepare work to modify null encode in index.
1. forbid range scan of bool in lookup, because `IndexColumnHint` can not make range [start, end) work.
2. add more test case for index scan rule test

### DIFF
--- a/src/optimizer/OptimizerUtils.cpp
+++ b/src/optimizer/OptimizerUtils.cpp
@@ -333,9 +333,6 @@ Value OptimizerUtils::boundValueWithMax(const meta::cpp2::ColumnDef& col) {
         case Value::Type::FLOAT : {
             return Value(std::numeric_limits<double>::max());
         }
-        case Value::Type::BOOL: {
-            return Value(true);
-        }
         case Value::Type::STRING : {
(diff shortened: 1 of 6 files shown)


src/optimizer/OptimizerUtils.cpp
```diff
@@ -338,7 +334,7 @@ Value OptimizerUtils::boundValueWithMax(const meta::cpp2::ColumnDef& col, const
             return Value(std::numeric_limits<double>::max());
         }
         case Value::Type::BOOL: {
-            return v;
+            return Value(true);
```
bright-starry-sky, line 337: I think , better move the BOOL to below of switch block. let's unify that with boundValueWithLT and boundValueWithGT . WDYT?

bright-starry-sky:
Good job. Thanks take this.

src/optimizer/OptimizerUtils.cpp
critical27: ok, make sense

bright-starry-sky (approved):
Well done.

dutor (approved):
Great

PR #426, opened 2020-11-21
Check remove host no duplicate

darionyaphet:


yixinglu (approved):

src/validator/BalanceValidator.cpp
```diff
@@ -19,13 +19,20 @@ Status BalanceValidator::toPlan() {
...
         current = BalanceLeaders::make(qctx_, nullptr);
         break;
-    case BalanceSentence::SubType::kData:
+    case BalanceSentence::SubType::kData: {
+        auto hosts = sentence->hostDel() == nullptr ? std::vector<HostAddr>()
+                                                    : sentence->hostDel()->hosts();
+        if (!hosts.empty()) {
+            auto it = std::unique(hosts.begin(), hosts.end());
+            if (it != hosts.end()) {
+                return Status::SemanticError("Host have duplicated");
```
laura-ding, line 28: I think it no need to return an error for the user.

dutor: > I think it no need to return an error for the user.

Whether to indicate an error depends on the impact the duplication may introduce.

dutor (approved):

PR #429, opened 2020-11-24
Modify status to stats

laura-ding:
depends on https://github.com/vesoft-inc/nebula-common/pull/320


panda-sheep (commented):
also  move `submit job statis`   to  `submit job stats`?

yixinglu (approved):

panda-sheep (approved):
Good, thx

PR #430, opened 2020-11-24
toss

liuyu85cn:
if space set toss enabled, add & update will only sent to out-edge.


src/parser/test/ParserTest.cpp
```diff
@@ -227,6 +227,13 @@ TEST(Parser, SpaceOperation) {
         auto result = parser.parse(query);
         ASSERT_TRUE(result.ok()) << result.status();
     }
+    {
+        GQLParser parser;
+        std::string query = "CREATE SPACE default_space(partition_num=9, replica_factor=3,"
+                            "atomic_edge=1)";
```
critical27, line 233: Maybe use bool will be better.

critical27 (dismissed):

critical27 (approved):

bright-starry-sky (approved):
LGTM

critical27 (approved):

src/parser/test/ParserTest.cpp
```diff
@@ -227,6 +227,20 @@ TEST(Parser, SpaceOperation) {
         auto result = parser.parse(query);
         ASSERT_TRUE(result.ok()) << result.status();
     }
+    {
+        GQLParser parser;
+        std::string query = "CREATE SPACE default_space(partition_num=9, replica_factor=3,"
+                            "atomic_edge=true)";
```
yixinglu, line 233: Isn't the transaction on edge operation supported by default? Why need user to specify it when creating space?

liuyu85cn: txn will disable by default (because of performance?)

dutor: I have the same doubts. I think the consistency guarantee is a MUST we have to provide. Otherwise, it will result in dirty data.

darionyaphet: seems we should ensure the transaction when writing in and out edges ?

darionyaphet (approved):
Like it ~

src/parser/test/ParserTest.cpp
critical27: Previously data is not guaranteed to be consistent, so as for use update from 1.0, enable toss by default can't fix previous inconsistency. If a new space, enable toss sounds reasonable.

On the other hand, we'd better test the whole 2.0 system when toss enabled, both for stability and performance before it is released officially. That will be a good time to enable it by default.

CPWstatic:
About the parameter name `atomic_edge`, why not name it `toss` directly?

liuyu85cn:
> About the parameter name `atomic_edge`, why not name it `toss` directly?

I thought most of the user will not understand what 'toss' refers to. 

BTW, I just look the dict, 'toss' is a really Eng word, which means 'throw' ...

darionyaphet:
> About the parameter name `atomic_edge`, why not name it `toss` directly?

Do you know what the full name of toss? It's too hard for the users ...

src/parser/test/ParserTest.cpp
dutor: > enable toss by default can't fix previous inconsistency

We can do nothing about the old dirty data anyway while it does not affect the fact that we _enable_ the so called toss by default. Thus I think it's even not necessary to introduce such an option.

critical27: > Thus I think it's even not necessary to introduce such an option.

Agree. After it is stable, we could remove it.

dutor: Fair enough

yixinglu:
> > About the parameter name `atomic_edge`, why not name it `toss` directly?
> 
> Do you know what the full name of toss? It's too hard for the users ...

`atomic_edge` also make users feel confusing.

critical27 (approved):

panda-sheep (approved):
LGTM

darionyaphet (approved):

PR #439, opened 2020-11-30
update docker image when new commits pushed to released branch

jievince:


.github/workflows/docker.yml
```diff
@@ -1,6 +1,9 @@
 name: docker
 
 on:
```
Shylock-Hg, line 3: The release job had do this.

yixinglu: yeah. This PR will solve the issue which we created a version branch and push new commits to this branch before releasing.

yixinglu (approved):
LGTM

.github/workflows/docker.yml
jievince: It's triggered after the release branch is created and before the released tag is created. During this time, if new commits pushed to the released branch, the docker image should be updated in time.

dutor (approved):

PR #447, opened 2020-12-01
Replace behave framework with pytest-bdd and add DataSetComparator

yixinglu:
Depends on nebula-python ValueWrapper compare implementation.


tests/tck/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+from tests.tck.utils.nbv import parse
+
+
+def _table(text):
+    lines = text.splitlines()
+    assert len(lines) >= 1
+
+    def parse_line(line):
+        return list(map(lambda x: x.strip(),
+                        filter(lambda x: x, line.split('|'))))
```
dutor, line 19: This will filter out the empty cells, right? Is that as expected?

tests/tck/utils/comparator.py
```diff
@@ -0,0 +1,152 @@
...
+        if self._order:
+            return all(self.compare_row(l, r) for (l, r) in zip(lhs, rhs))
+        return self._compare_list(lhs, rhs, self.compare_row)
+
+    def compare_value(self, lhs: ValueWrapper, rhs: ValueWrapper):
+        '''
+        lhs and rhs represent response data and expected data respectively
+        '''
+        if lhs.is_null():
+            return rhs.is_null()
```
dutor, line 40: Shall we elaborate the individual null types?

dutor (commented):
Excellent

tests/tck/conftest.py
yixinglu: yeah. We don't allow following case `| nothing here |` in expected table, so is safe to filter the empty cells.

tests/tck/utils/comparator.py
yixinglu: Maybe. But we can let it go at present. When we encounter those test cases, it's not late to fix it. Will add some todo comments.

ci/test.sh
```diff
@@ -91,16 +91,12 @@ function run_ctest() {
...
-        -n=8 \
+    pytest -n 8 --build_dir=$BUILD_DIR \
         --dist=loadfile \
         --debug_log=false \
-        ${@:1} \
-        $testpath
+        ${@:1}
 
-    $BUILD_DIR/tests/ntr --debug_log=false ${@:1} $PROJ_DIR/tests/job/*
+    # $BUILD_DIR/tests/ntr --debug_log=false ${@:1} $PROJ_DIR/tests/job/*
```
laura-ding, line 99: Why comment it out

tests/tck/utils/comparator.py
```diff
@@ -0,0 +1,148 @@
...
+            and self.compare_node(l.end_node, r.end_node)
+            and self.compare_edge(l.relationship, r.relationship)
+            for (l, r) in zip(lhs, rhs))
+
+    def compare_edge(self, lhs: Relationship, rhs: Relationship):
+        if not lhs == rhs:
+            return False
+        if not self._strict:
+            return True
+        return self.compare_map(lhs.propertys(), rhs.propertys())
```
laura-ding, line 95: need to compare edge key

tests/tck/utils/comparator.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def compare_node(self, lhs: Node, rhs: Node):
+        if lhs.get_id() != rhs.get_id():
+            return False
+        if not self._strict:
+            return True
+        for tag in lhs.tags():
```
laura-ding, line 102: check the size of the tags firstly

tests/tck/utils/comparator.py
```diff
@@ -0,0 +1,148 @@
...
+        if lhs.is_edge():
+            return rhs.is_edge() and self.compare_edge(lhs.as_relationship(),
+                                                       rhs.as_relationship())
+        if lhs.is_path():
+            return rhs.is_path() and self.compare_path(lhs.as_path(),
+                                                       rhs.as_path())
+        return False
+
+    def compare_path(self, lhs: PathWrapper, rhs: PathWrapper):
+        return all(
```
laura-ding, line 84: check the path size firstly

tests/tck/utils/comparator.py
yixinglu: nebula python client has compared edge keys in `Relationship::__eq__`. `lhs == rhs` will call the equal function.

tests/tck/utils/comparator.py
yixinglu: fixed it.

tests/tck/utils/comparator.py
yixinglu: fixed it.

ci/test.sh
yixinglu: will fixed job test cases in next PR

yixinglu: will fix job test cases in next PR

dutor (approved):

tests/conftest.py
```diff
@@ -62,6 +73,184 @@ def pytest_configure(config):
...
+            load_csv_data(pytestconfig, conn_pool, "student")
+            fn.write_text("student")
+            logging.info(f"session-{worker_id} load student csv data")
+            load = True
+        else:
+            logging.info(
+                f"session-{worker_id} need not to load student csv data")
+    yield
+    if load:
+        os.remove(str(fn))
```
laura-ding, line 216: you can define one function, and make the space_name as input args

tests/query/v1/test_find_path.py
```diff
@@ -589,7 +589,7 @@ def test_path_limit(self):
             ]
         }
         self.check_column_names(resp, expected_data["column_names"])
-        self.check_path_result_without_prop(resp.rows(), expected_data["rows"])
+        # self.check_path_result_without_prop(resp.rows(), expected_data["rows"])
```
laura-ding, line 592: why comment?

laura-ding (approved):
excellent! please modify the readme for the tests

tests/query/v1/test_find_path.py
yixinglu: will be fixed in next PR.

tests/conftest.py
yixinglu: make sense, I am thinking how to optimize these fixtures if the `yield`sentence in different functions.

PR #454, opened 2020-12-03
fix ErrorCode typo

Aiee:
depends on https://github.com/vesoft-inc/nebula-common/pull/333

### DIFF
--- a/src/service/GraphService.cpp
+++ b/src/service/GraphService.cpp
@@ -99,9 +99,8 @@ const char* GraphService::getErrorStr(ErrorCode result) {
         return "Syntax error";
     case ErrorCode::E_SEMANTIC_ERROR:
         return "Semantic error";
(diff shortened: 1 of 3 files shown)


src/service/GraphService.cpp
```diff
@@ -100,8 +100,8 @@ const char* GraphService::getErrorStr(ErrorCode result) {
     case ErrorCode::E_SEMANTIC_ERROR:
         return "Semantic error";
     // TODO(shylock) fix the typo
```
Shylock-Hg, line 102: Remove the TODO comment.

Aiee: fixed

Shylock-Hg (approved):

laura-ding (approved):

laura-ding:
need to update python client thrift files

PR #467, opened 2020-12-10
Support predicate expression.

jievince:
depends on https://github.com/vesoft-inc/nebula-common/pull/345
doc:https://neo4j.com/docs/cypher-manual/current/functions/predicate/


src/util/ParserUtil.h
```diff
@@ -0,0 +1,75 @@
...
+                    ret = la->clone().release();
+                }
+            }
+            return ret;
+        };
+
+        RewriteMatchLabelVisitor visitor(rewriter);
+
+        pred->setOriginString(new std::string(pred->makeString()));
+        pred->setInnerVar(new std::string(newVarName));
```
CPWstatic, line 60: I think we should not simply replace the original var name, or `pred ->toString()` will not give a proper string of the PredicateExpression.

CPWstatic (dismissed):
LGTM.

src/util/ParserUtil.h
jievince: I have use method `setOriginString()` to save the origin string of predicate expression before it is rewritten, so `pred->toString()` will show the origin string properly:
```
std::string PredicateExpression::toString() const {
    if (originString_ != nullptr) {
        return *originString_;
    } else {
        return makeString();
    }
}

CPWstatic (approved):

yixinglu (approved):

yixinglu (approved):

CPWstatic (approved):

PR #470, opened 2020-12-11
Add session 0 processing

laura-ding:


jievince (approved):

Aiee (approved):

src/service/GraphService.cpp
```diff
@@ -66,11 +66,18 @@ GraphService::future_execute(int64_t sessionId, const std::string& query) {
     ctx->setRunner(getThreadManager());
     auto future = ctx->future();
     {
+        // When the sessionId is 0, it means the clients to ping the connection is ok
+        if (sessionId == 0) {
+            ctx->resp().errorCode = ErrorCode::E_SESSION_INVALID;
+            ctx->resp().errorMsg = std::make_unique<std::string>("The Session id is not valid");
```
CPWstatic, line 72: Should we add some log here?

laura-ding: it's in order to not print logs

PR #480, opened 2020-12-14
Fix some issue about lookup the yield and the outputs.

Shylock-Hg:


src/validator/LookupValidator.h
```diff
@@ -68,7 +68,11 @@ class IndexScanValidator final : public Validator {
     bool                              textSearchReady_{false};
     std::string                       from_;
     std::vector<nebula::plugin::HttpClient> esClients_;
+    std::vector<std::string>          isColNames_;
```
yixinglu, line 71: Maybe `idxScanColNames_` is better

tests/query/stateless/test_lookup.py
```diff
@@ -127,33 +102,113 @@ def test_tag_index(self):
         resp = self.execute('INSERT VERTEX team(name) VALUES "204":("opl")')
         self.check_resp_succeeded(resp)
 
+
+    def test_edge_index(self):
```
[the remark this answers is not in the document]
Shylock-Hg: These case insert the new data, I plan to rewrite them with the exist data like `nba`.

src/validator/LookupValidator.h
Shylock-Hg: Fixed.

src/validator/LookupValidator.cpp
```diff
@@ -14,36 +14,55 @@ DECLARE_uint32(ft_request_retry_times);
...
+
+    if (dedup_) {
+        auto* dedupNode = Dedup::make(qctx_, current);
+        dedupNode->setInputVar(current->outputVar());
+        dedupNode->setColNames(colNames_);
+        current = dedupNode;
+
+        // the framework will add data collect to collect the result
+        // if the result is required
+    }
```
bright-starry-sky, line 57: a question about which fields need to be de-dup？any field？

Shylock-Hg: The row in result.

bright-starry-sky: > The row in result.

got, If it's just rows de-dup, the graph layer doesn't need a separate node, a bool parameter is enough, It will be pushed down to the storage layer.
If any fields, the deduce node is required. for example :
lookup on t1 where ... yield distinct t1.c1, t1.c2
lookup on t1 where ... yield distinct _src, _dst

So we need to think about final design. That's why I want to remove `idxScanColNames_` of duplicate variable.

CPWstatic (approved):
LGTM. But don't try to format the irrelative codes.

tests/tck/features/lookup/Output.feature
```diff
@@ -0,0 +1,96 @@
...
+    When executing query:
+      """
+      $a = LOOKUP ON serve WHERE serve.start_year == 2008 and serve.end_year == 2019
+      YIELD serve.start_year AS startYear;
+      FETCH PROP ON serve $a.SrcVID->$a.DstVID YIELD serve.start_year AS startYear
+      """
+    Then the result should be, in any order:
+      | serve._src          | serve._dst  | serve._rank | startYear |
+      | 'Russell Westbrook' | 'Thunders'  | 0           | 2008      |
+      | 'Marc Gasol'        | 'Grizzlies' | 0           | 2008      |
```
laura-ding, line 96: You can add some test to fail

src/validator/LookupValidator.cpp
```diff
@@ -55,72 +74,100 @@ Status IndexScanValidator::prepareFrom() {
...
+                    new std::string(schemaName), new std::string(colName))));
+            } else {
+                newYieldColumns_->addColumn(new YieldColumn(new TagPropertyExpression(
+                    new std::string(schemaName), new std::string(colName))));
+            }
+            if (col->alias() != nullptr) {
+                newYieldColumns_->back()->setAlias(new std::string(*col->alias()));
+            }
+            if (schemaName != from_) {
+                return Status::SemanticError("Schema name error : %s", schemaName.c_str());
```
laura-ding, line 145: here is better tell user the `col->expr().toString()` is wrong.

laura-ding (approved):

PR #488, opened 2020-12-16
Support int vid in mutate sentences

laura-ding:
Test with the pytest. here not add the test file.
TODO: need to write features file,  but the features file need another pr supported
https://github.com/vesoft-inc/nebula-graph/pull/475


src/parser/parser.yy
```diff
@@ -996,9 +999,6 @@ vid
         $$ = new ConstantExpression(*$1);
         delete $1;
     }
-    | legal_integer {
```
Shylock-Hg, line 999: Why change this?

laura-ding: `legal_integer` is a positive integer, `int vid` can be a negative integer.

Shylock-Hg: Ok.

nevermore3 (approved):

nevermore3 (approved):

CPWstatic (approved):

yixinglu:
I don't think it's an improvement about `addDescription` changes. At most scenarios, we don't pass a `Value` to it, it's not necessary to do this conversion from string to Value and from Value to string again.

PR #517, opened 2020-12-24
Fix the bug of datetime default value eval in metad which don't initi…

Shylock-Hg:
…alize the timezone.

Require https://github.com/vesoft-inc/nebula-storage/pull/262

NG-351


yixinglu (commented):
shall we test vertex insertion with these tags which datetime properties are not set?

Shylock-Hg:
> shall we test vertex insertion with these tags which datetime properties are not set?

The other case had contains the crud.

yixinglu (approved):

CPWstatic (approved):

PR #522, opened 2020-12-24
Fix same column expect result issue

yixinglu:
as title.


CPWstatic (approved):

tests/tck/features/subgraph/subgraph.IntVid.feature
```diff
@@ -4,7 +4,7 @@
 # attached with Common Clause Condition 1.0, found in the LICENSES directory.
 Feature: Integer Vid subgraph
 
-  Background: Prepare space
```
Shylock-Hg, line 7: Why change this?

yixinglu: I lookup the docs of pytest-bdd, it's not necessary to add this description in background.

nevermore3 (approved):

PR #539, opened 2020-12-29
add 1.x case vid int case of lookup and index 

tom-chensf:


bright-starry-sky:
Good job. what’s progress? any issue please let me know.

CPWstatic (approved):

laura-ding (commented):
When Scenario contains multiple test cases, it is better to annotate each test case to illustrate the test Scenario.
Now here is so many skip test.

PR #549, opened 2020-12-31
Fix GetNeighborsExecutor empty input result

laura-ding:


src/executor/query/GetNeighborsExecutor.cpp
```diff
@@ -69,7 +69,7 @@ Status GetNeighborsExecutor::buildRequestDataSet() {
 folly::Future<Status> GetNeighborsExecutor::getNeighbors() {
     if (reqDs_.rows.empty()) {
         LOG(INFO) << "Empty input.";
-        DataSet emptyResult;
```
yixinglu, line 72: Executor output result should always be a `DataSet`. what problem does the empty DataSet have?

laura-ding: The GetNeighborsIter only accept the List < DataSet >

src/executor/query/GetNeighborsExecutor.cpp
```diff
@@ -69,7 +69,7 @@ Status GetNeighborsExecutor::buildRequestDataSet() {
 folly::Future<Status> GetNeighborsExecutor::getNeighbors() {
     if (reqDs_.rows.empty()) {
         LOG(INFO) << "Empty input.";
```
CPWstatic, line 71: Could we change the log here to VLOG?

yixinglu (approved):

CPWstatic (approved):

PR #550, opened 2020-12-31
Erase by swap in filter.

Shylock-Hg:
From O(N^2) to O(N) but lost the order of elements (which we don't need in the modified executor in fact).


nevermore3 (approved):

yixinglu (dismissed):

src/context/test/IteratorTest.cpp
```diff
@@ -399,6 +399,66 @@ TEST(IteratorTest, GetNeighbor) {
...
         EXPECT_EQ(result.values, expected);
     }
+    // unstableErase
+    {
+        GetNeighborsIter iter(val);
+        size_t i = 0;
+        while (iter.valid()) {
+            ++i;
+            if (i % 2 == 0) {
+                iter.erase();
```
yixinglu, line 409: unstableErase() ?

Shylock-Hg: Good catch. Fixed.

src/context/Iterator.h
```diff
@@ -179,6 +191,11 @@ class DefaultIter final : public Iterator {
         counter_--;
     }
 
+    void unstableErase() override {
+        DLOG(ERROR) << "Unimplemented default iterator.";
```
CPWstatic, line 195: Would it be better if we use DCHECK here?

CPWstatic (approved):
LGTM.

CPWstatic:
Emmm, we need implement the move constructor for logicalrows.

PR #556, opened 2021-01-04
Add flag to determine whether partial success results are valid

Aiee:
As title

depends on https://github.com/vesoft-inc/nebula-common/pull/377

### DIFF
--- a/conf/nebula-graphd.conf.default
+++ b/conf/nebula-graphd.conf.default
@@ -24,7 +24,8 @@
 --stderrthreshold=2
 
 ########## query ##########
-# Whether to treat partial success as an error
+# Whether to treat partial success as an error. 
+# This flag is only used for Read-only access, and Modify access always treats partial success as an error.
 --accept_partial_success=false
 
 ########## networking ##########
--- a/conf/nebula-graphd.conf.default
+++ b/conf/nebula-graphd.conf.default
@@ -24,7 +24,8 @@
 --stderrthreshold=2
 
 ########## query ##########
---complete_require=true
+# Whether to treat partial success as an error
+--accept_partial_success=false
 
 ########## networking ##########
 # Comma separated Meta Server Addresses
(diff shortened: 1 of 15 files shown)


src/executor/query/GetNeighborsExecutor.h
```diff
@@ -17,6 +17,7 @@
 
 #include "executor/QueryStorageExecutor.h"
 #include "planner/Query.h"
+#include "service/GraphFlags.h"
```
yixinglu, line 20: Need not to include this header file here.

src/executor/query/IndexScanExecutor.h
```diff
@@ -11,6 +11,7 @@
 #include "common/clients/storage/GraphStorageClient.h"
 #include "executor/QueryStorageExecutor.h"
 #include "planner/Query.h"
+#include "service/GraphFlags.h"
```
yixinglu, line 14: Move this include stmt to cpp file.

conf/nebula-graphd.conf.default
```diff
@@ -23,6 +23,10 @@
 # Copy log messages at or above this level to stderr in addition to logfiles. The numbers of severity levels INFO, WARNING, ERROR, and FATAL are 0, 1, 2, and 3, respectively.
 --stderrthreshold=2
 
+########## query ##########
+# Whether to treat partial success as an error
+--accept_partial_success=false
```
Shylock-Hg, line 28: Comment that it's only used for read-only access.

Shylock-Hg: Modify access don't accept partial success.

Aiee: Comments have been updated

src/service/QueryInstance.cpp
```diff
@@ -142,6 +142,7 @@ void QueryInstance::onError(Status status) {
         case Status::Code::kIndexNotFound:
         case Status::Code::kInserted:
         case Status::Code::kKeyNotFound:
+        case Status::Code::kPartialSuccess:
```
Shylock-Hg, line 145: When return this error?

Aiee: Currently this error will be processed as ErrorCode::E_EXECUTION_ERROR in QueryInstance

yixinglu (approved):

Shylock-Hg (approved):

PR #558, opened 2021-01-05
add assign executor

nevermore3:
Support assigning values ​​to variables during execution


CPWstatic (approved):

Shylock-Hg:
What's the purpose?

nevermore3:
> What's the purpose?

When the loop is nested inside the loop, the nested loop cannot be initialized every time
for example, execution plan is:
    Loop1-> ...  ->Loop2
can be convert to:
 for i = 0; i < N; ++i
         for j = 0; j < i; j++
         done
 done 
Every time the outer loop, the inner loop  j should be initialized

yixinglu (approved):

PR #559, opened 2021-01-05
Select the tag index by key length.

Shylock-Hg:
Require https://github.com/vesoft-inc/nebula-common/pull/378

### DIFF
--- a/src/context/test/CMakeLists.txt
+++ b/src/context/test/CMakeLists.txt
@@ -28,6 +28,7 @@ SET(CONTEXT_TEST_LIBS
     $<TARGET_OBJECTS:common_time_utils_obj>
     $<TARGET_OBJECTS:common_graph_obj>
     $<TARGET_OBJECTS:common_ft_es_graph_adapter_obj>
+    $<TARGET_OBJECTS:common_index_utils_obj>
     $<TARGET_OBJECTS:util_obj>
     $<TARGET_OBJECTS:context_obj>
     $<TARGET_OBJECTS:expr_visitor_obj>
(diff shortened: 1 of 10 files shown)


bright-starry-sky:
I don't think this is the right way to handle it. I think choosing an optimal index is based on where conditions, not fields size. 
Of course, we can sort the indices and then make a choice. 
Is a BUG in your PR . for example :
```
tag1 (col1 string)
index1 on tag (col1(2))
index2 on tag (col1(5))
where col1 == "AA" , --> The index1 will be hit. The choice is correct
where col1 == "AAAAA" -->The index1 will be hit too. The choice is wrong. index2 is expected.
```

WDYT?

Shylock-Hg:
> I don't think this is the right way to handle it. I think choosing an optimal index is based on where conditions, not fields size.
> Of course, we can sort the indices and then make a choice.
> Is a BUG in your PR ? for example :
> 
> ```
> tag1 (col1 string)
> index1 on tag (col1(2))
> index2 on tag (col1(5))
> where col1 == "AA" , --> The index1 will be hit. The choice is correct
> where col1 == "AAAAA" -->The index1 will be hit too. The choice is wrong. index2 is expected.
> ```
> 
> WDYT?

No filter if we seek by index itself as the start of match.

bright-starry-sky:
> > I don't think this is the right way to handle it. I think choosing an optimal index is based on where conditions, not fields size.
> > Of course, we can sort the indices and then make a choice.
> > Is a BUG in your PR ? for example :
> > ```
> > tag1 (col1 string)
> > index1 on tag (col1(2))
> > index2 on tag (col1(5))
> > where col1 == "AA" , --> The index1 will be hit. The choice is correct
> > where col1 == "AAAAA" -->The index1 will be hit too. The choice is wrong. index2 is expected.
> > ```
> > 
> > 
> > WDYT?
> 
> No filter if we seek by index itself as the start of match.

Yes, I know. I means LOOKUP and MATCH should have the same processing logic, don't do it this way, otherwise it will be more cluttered.
And the processing logic, I suggest you look at` count (vertices)` by tag.

Shylock-Hg:
Stale

PR #569, opened 2021-01-08
check uniqueness of yield column alias

jievince:


tests/tck/features/match/Base.feature
```diff
@@ -388,3 +388,34 @@ Feature: Basic match
       MATCH () --> (v) --> () return *
       """
     Then a ExecutionError should be raised at runtime: Can't solve the start vids from the sentence: MATCH ()-->(v)-->() RETURN *
+
+  Scenario: Return columns with same name
```
Shylock-Hg, line 392: We only check it when output to successor sentence.

jievince: But `RETURN` is the last clause.

Shylock-Hg: I means you don't need check the output, the others sentence will check the input from match.

PR #585, opened 2021-01-11
fix getNeighbor iterator

nevermore3:
Add rows without edge to noEdgeRows_ , 
Then Subgraph can collect the vertices of each step for the getNeigbors


nevermore3:
D8B4924B-D10E-4171-9069-7E8ED5AAF055
we should add  (Pacers) and (Thunders) to logicalRow

src/context/Iterator.cpp
```diff
@@ -96,6 +96,7 @@ StatusOr<GetNeighborsIter::DataSetIndex> GetNeighborsIter::makeDataSetIndex(cons
 void GetNeighborsIter::makeLogicalRowByEdge(int64_t edgeStartIndex,
                                             size_t idx,
                                             const DataSetIndex& dsIndex) {
+    bool existEdge = false;
     for (auto& row : dsIndex.ds->rows) {
         auto& cols = row.values;
```
yixinglu, line 101: Init `existEdge` variable here

src/context/Iterator.cpp
```diff
@@ -108,12 +109,17 @@ void GetNeighborsIter::makeLogicalRowByEdge(int64_t edgeStartIndex,
...
                 auto edgeName = dsIndex.tagEdgeNameIndices.find(column);
                 DCHECK(edgeName != dsIndex.tagEdgeNameIndices.end());
                 logicalRows_.emplace_back(
                     idx, &row, edgeName->second, &edge.getList());
             }
         }
+        if (!existEdge) {
+            logicalRows_.emplace_back(idx, &row, "", nullptr);
+        }
+        existEdge = false;
```
yixinglu, line 122: And need not it

yixinglu (commented):
Could you add some test cases for this scenario ?

src/context/Iterator.cpp
nevermore3: done

src/context/Iterator.cpp
nevermore3: done

nevermore3:
> Could you add some test cases for this scenario ?

done

CPWstatic (approved):

Shylock-Hg (approved):

yixinglu (approved):

PR #591, opened 2021-01-12
Make StartNode first when generating Execution Plan

jievince:


src/planner/PlanNode.h
```diff
@@ -232,6 +232,7 @@ class PlanNode {
 
     void clone(const PlanNode &node) {
         // TODO maybe shall copy cost_ and dependencies_ too
+        id_ = node.id_;
```
yixinglu, line 235: how to differentiate these two plan nodes by id if they have same node ids?

jievince: My fault...Has fixed

nevermore3 (approved):

Aiee (approved):

PR #611, opened 2021-01-14
[Simple] Removed ConfigNebulaStorage.cmake

sherman-the-tank:
depends on vesoft-inc/nebula-common#387


.gitignore
```diff
@@ -9,6 +9,7 @@ _build.log
 _install
 install
 bin/
+tests/
```
yixinglu, line 12: should not ignore `tests` directory since our tests are in that directory.

sherman-the-tank: Good catch 👍

yixinglu (approved):

PR #619, opened 2021-01-14
Fix bug about match index selection

yixinglu:
Solved following two bugs which result in performance regression in MATCH:

```
MATCH (v)--(v2) WHERE id(v2)=="xx" RETURN v2

MATCH (v:player{name: "xx"}) RETURN v;  -- Could not select a right index and scan all tags
```

There're still some issues for match, I will refactor the index selection logic in next PR.

reference https://discuss.nebula-graph.com.cn/t/topic/2397/13


Shylock-Hg:
The case 1 had fixed in https://github.com/vesoft-inc/nebula-graph/pull/500

yixinglu:
> The case 1 had fixed in #500

We can discuss the better solution of your PR offline. This is a hotfix.

jievince (approved):

nevermore3 (approved):

PR #625, opened 2021-01-15
Adapt the max return edge.

CPWstatic:
After https://github.com/vesoft-inc/nebula-storage/pull/300


src/planner/Query.h
```diff
@@ -133,7 +133,7 @@ class GetNeighbors final : public Explore {
                               bool dedup = false,
                               bool random = false,
                               std::vector<storage::cpp2::OrderBy> orderBy = {},
-                              int64_t limit = std::numeric_limits<int64_t>::max(),
```
Shylock-Hg, line 136: What's the problem?

panda-sheep (approved):
LGTM

yixinglu (approved):

PR #636, opened 2021-01-18
Use exact regular matching in nbv.py

jievince:


yixinglu:
@dutor Could you fix it later? I close this PR now.

PR #669, opened 2021-01-21
Add opt rules to delete redundant dedup and project plan node

yixinglu:
1. Merge GetVertices and Dedup: GetVertices(dedup=true)
2. Merge GetNeighbors and Dedup: GetNeighbors(dedup=true)
3. Merge GetVertices and Project: GetVertices(srcExpr=projectExpr)
4. Merge GetNeighbors and Project: GetNeighbors(srcExpr=projectExpr)

depends on #679

### DIFF
--- a/src/executor/query/GetVerticesExecutor.cpp
+++ b/src/executor/query/GetVerticesExecutor.cpp
@@ -26,23 +26,28 @@ folly::Future<Status> GetVerticesExecutor::getVertices() {
     SCOPED_TIMER(&execTime_);
 
     auto *gv = asNode<GetVertices>(node());
-
     GraphStorageClient *storageClient = qctx()->getStorageClient();
     nebula::DataSet vertices({kVid});
-    const auto& spaceInfo = qctx()->rctx()->session()->space();
     if (gv->src() != nullptr) {
         // Accept Table such as | $a | $b | $c |... as input which one column indicate src
         auto valueIter = ectx_->getResult(gv->inputVar()).iter();
-        VLOG(1) << "GV input var: " << gv->inputVar() << " iter kind: " << valueIter->kind();
+        VLOG(3) << "GV input var: " << gv->inputVar() << " iter kind: " << valueIter->kind();
         auto expCtx = QueryExpressionContext(qctx()->ectx());
+        const auto &spaceInfo = qctx()->rctx()->session()->space();
+        std::unordered_set<Value> uniqueSet;
         for (; valueIter->valid(); valueIter->next()) {
             auto src = gv->src()->eval(expCtx(valueIter.get()));
-            VLOG(1) << "src vid: " << src;
             if (!SchemaUtil::isValidVid(src, spaceInfo.spaceDesc.vid_type)) {
                 LOG(WARNING) << "Mismatched vid type: " << src.type();
                 continue;
             }
-            vertices.emplace_back(Row({std::move(src)}));
+            if (gv->dedup()) {
+                if (uniqueSet.emplace(src).second) {
+                    vertices.emplace_back(Row({std::move(src)}));
+                }
+            } else {
+                vertices.emplace_back(Row({std::move(src)}));
+            }
         }
     }
 
--- a/src/optimizer/rule/MergeGetNbrsAndDedupRule.cpp
+++ b/src/optimizer/rule/MergeGetNbrsAndDedupRule.cpp
@@ -6,6 +6,7 @@
 
 #include "optimizer/rule/MergeGetNbrsAndDedupRule.h"
 
+#include "optimizer/OptContext.h"
 #include "optimizer/OptGroup.h"
 #include "planner/PlanNode.h"
 #include "planner/Query.h"
@@ -13,7 +14,6 @@
 using nebula::graph::Dedup;
 using nebula::graph::GetNeighbors;
 using nebula::graph::PlanNode;
-using nebula::graph::QueryContext;
 
 namespace nebula {
 namespace opt {
@@ -32,7 +32,7 @@ const Pattern &MergeGetNbrsAndDedupRule::pattern() const {
 }
 
 StatusOr<OptRule::TransformResult> MergeGetNbrsAndDedupRule::transform(
-    QueryContext *qctx,
+    OptContext *ctx,
     const MatchedResult &matched) const {
     const OptGroupNode *optGN = matched.node;
     const OptGroupNode *optDedup = matched.dependencies.back().node;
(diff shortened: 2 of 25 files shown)


jievince (approved):
LGTM

src/optimizer/rule/MergeGetNbrsAndDedupRule.cpp
```diff
@@ -0,0 +1,63 @@
...
+    DCHECK_EQ(optDedup->node()->kind(), PlanNode::Kind::kDedup);
+    auto gn = static_cast<const GetNeighbors *>(optGN->node());
+    auto dedup = static_cast<const Dedup *>(optDedup->node());
+    auto newGN = gn->clone(qctx);
+    if (!newGN->dedup()) {
+        newGN->setDedup();
+    }
+    newGN->setInputVar(dedup->inputVar());
+    auto newOptGV = OptGroupNode::create(qctx, newGN, optGN->group());
+    for (auto dep : optDedup->dependencies()) {
```
Shylock-Hg, line 49: In fact I think this dedup node is still necessary (e.g. request to multiple hosts that the push-down dedup option only dedup result from one host).

src/optimizer/rule/MergeGetVerticesAndDedupRule.cpp
```diff
@@ -0,0 +1,63 @@
...
+    auto gv = static_cast<const GetVertices *>(optGV->node());
+    auto dedup = static_cast<const Dedup *>(optDedup->node());
+    auto newGV = gv->clone();
+    if (!newGV->dedup()) {
+        newGV->setDedup();
+    }
+    newGV->setInputVar(dedup->inputVar());
+    auto newOptGV = OptGroupNode::create(qctx, newGV, optGV->group());
+    for (auto dep : optDedup->dependencies()) {
+        newOptGV->dependsOn(dep);
```
Shylock-Hg, line 50: ditto.

Shylock-Hg:
And another thing is that the dependencies it the execution relationship not the input/output relationship, for the dedup depends on the gv/gn maybe not get input from them

src/optimizer/rule/MergeGetNbrsAndDedupRule.cpp
yixinglu: The removed dedup node is the input of GetNeighbors, rather than dedup the result of GetNeighbors. The execution plan is like:

```text
...-> Dedup -> (vids) -> GetNeighbors -> ...

=>

...-> GetNeighbors -> ...  # dedup vids inside GetNeighbors
```

The benefits of the rule is avoid the redundant src expression evaluation.

Shylock-Hg: ok.

yixinglu:
> And another thing is that the dependencies means the execution relationship not the input/output relationship, for the dedup depends on the gv/gn maybe not get input from them

you are right. But here is not the scenario which dedup depends on gv/gn. It‘s the relationship that gn/gv depends on dedup. And we should not let some plan nodes depend on the node which don't materialize the data into a real dataset (like filter/dedup) in dataflow. I could check the dataflow dependencies, but i feel it's not necessary.

Shylock-Hg:
> > And another thing is that the dependencies means the execution relationship not the input/output relationship, for the dedup depends on the gv/gn maybe not get input from them
> 
> you are right. But here is not the scenario which dedup depends on gv/gn. It‘s the relationship that gn/gv depends on dedup. And we should not let some plan nodes depend on the node which don't materialize the data into a real dataset (like filter/dedup) in dataflow. I could check the dataflow dependencies, but i feel it's not necessary.

Yes , but better check the dedup output is the input of gv/gn

yixinglu:
> > > And another thing is that the dependencies means the execution relationship not the input/output relationship, for the dedup depends on the gv/gn maybe not get input from them
> > 
> > 
> > you are right. But here is not the scenario which dedup depends on gv/gn. It‘s the relationship that gn/gv depends on dedup. And we should not let some plan nodes depend on the node which don't materialize the data into a real dataset (like filter/dedup) in dataflow. I could check the dataflow dependencies, but i feel it's not necessary.
> 
> Yes , but better check the dedup output is the input of gv/gn

Fixed in #679

src/executor/query/GetVerticesExecutor.cpp
```diff
@@ -26,23 +27,28 @@ folly::Future<Status> GetVerticesExecutor::getVertices() {
...
+        std::unordered_set<Value> uniqueSet;
         for (; valueIter->valid(); valueIter->next()) {
             auto src = gv->src()->eval(expCtx(valueIter.get()));
-            VLOG(1) << "src vid: " << src;
             if (!SchemaUtil::isValidVid(src, spaceInfo.spaceDesc.vid_type)) {
                 LOG(WARNING) << "Mismatched vid type: " << src.type();
                 continue;
             }
-            vertices.emplace_back(Row({std::move(src)}));
+            if (gv->dedup()) {
```
Shylock-Hg, line 45: This dedup option is applied for the result not the input.But now we return vid always so it's workable. The default return column is plan to remove.

CPWstatic:
In most cases, we have such a plan:
```
gn->dedup->project
gv->dedup->project
```
So we could also merge the project too.

src/executor/query/GetVerticesExecutor.cpp
yixinglu: Is there some problems? If we dedup the input vid of vertices, GV always returns unique result set whether we pass the dedup option into the storage or not.

yixinglu:
> In most cases, we have such a plan:
> 
> ```
> gn->dedup->project
> gv->dedup->project
> ```
> 
> So we could also merge the project too.

This rule handle the cases: `dedup -> gn` or `dedup -> gv`. Maybe we should provide some new rules for your scenarios.

jievince (dismissed):
LGTM~

src/executor/query/GetVerticesExecutor.cpp
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "executor/query/GetVerticesExecutor.h"
+#include <unordered_set>
```
laura-ding, line 8: the `#include "common/base/Base.h"` has include it

laura-ding (dismissed):

yixinglu:
> In most cases, we have such a plan:
> 
> ```
> gn->dedup->project
> gv->dedup->project
> ```
> 
> So we could also merge the project too.

Good ideas, I will refactor this PR to optimize the pattern: gv/gn -> dedup -> project

yixinglu:
@CPWstatic I rethink the rule again, I prefer to provide another rule to merge project and gv/gn since sometimes there exists some subplan which composed of project and gv/gn. If so this new opt rule could optimize the pattern.

jievince (approved):

laura-ding (approved):

nevermore3 (approved):

PR #679, opened 2021-01-22
Check dataflow dependency between plan nodes in optimizer

yixinglu:
At present, we don't handle the plan node which depends on other node in data flow.

This PR provide an OptContext class which could free OptGroup/OptGroupNode memory early when finishing optimization. And the OptContext also maintain the relationship between OptGroupNode and PlanNode to check the dependencies of data flow.


jievince (dismissed):
LGTM

nevermore3 (dismissed):

src/optimizer/OptRule.cpp
```diff
@@ -69,10 +71,39 @@ StatusOr<MatchedResult> OptRule::match(const OptGroupNode *groupNode) const {
...
+    return checkDataflowDeps(matched, matched.node->node()->outputVar(), true);
+}
+
+bool OptRule::checkDataflowDeps(const MatchedResult &matched,
+                                const std::string &var,
+                                bool isRoot) const {
+    auto node = matched.node;
+    auto planNode = node->node();
+    const auto &outVarName = planNode->outputVar();
+    if (outVarName != var) {
```
laura-ding, line 83: the `outVarName ` is always equal to `var`, they are the same one.

yixinglu: Not always, this pr check whether the output variable of dependency is input variable of current plan node in line 101 and line 105.

jievince: > the `outVarName ` is always equal to `var`, they are the same one.

they are just the same in the first invoke.

jievince (approved):

src/optimizer/OptRule.cpp
```diff
@@ -69,10 +71,39 @@ StatusOr<MatchedResult> OptRule::match(const OptGroupNode *groupNode) const {
...
+        return false;
+    }
+
+    if (matched.dependencies.empty()) {
+        return true;
+    }
+
+    DCHECK_EQ(matched.dependencies.size(), node->dependencies().size());
+
+    if (matched.dependencies.size() == 1U) {
```
Shylock-Hg, line 99: I think you could call it recursively in for loop directly instead of check size of dependencies.

laura-ding (dismissed):

src/optimizer/OptRule.cpp
yixinglu: Good idea. I'll refactor the inputVar getter of PlanNode to do this.

Shylock-Hg (dismissed):

src/util/Utils.h
```diff
@@ -0,0 +1,31 @@
...
+#include <string>
+#include <vector>
+
+#include <folly/String.h>
+
+namespace nebula {
+namespace util {
+
+template <typename Container, typename Fn>
+std::string join(const Container& container, Fn fn, const std::string& delimiter = ",") {
```
Shylock-Hg, line 20: Could add a template specialization for vector itself to avoid extra copy.

yixinglu: I want to let the `join` function adapt to sequential container type, like `std::vector` or `std::list`. And the `container` parameter is a const reference which will not copy the real parameter value.

nevermore3 (approved):

jievince (approved):

PR #684, opened 2021-01-25
ut add path filter by dst node

jude-zhu:
from users


nevermore3 (approved):

randomJoe211 (approved):
👍 
I've updated the MATCH doc accordingly.

PR #685, opened 2021-01-25
fix go from id($-.vertices_)

jievince:
Currently parser couldn't distinguish between `hash("Tony Parker")`(constant vid) and `id($-.vertices_)`(runtime vid). It treats all function_call_expression as constant vid.


src/visitor/FindAnyExprVisitor.h
```diff
@@ -17,7 +17,7 @@ namespace graph {
 
 class FindAnyExprVisitor final : public ExprVisitorImpl {
 public:
-    explicit FindAnyExprVisitor(const std::unordered_set<Expression::Kind>& kinds);
+    explicit FindAnyExprVisitor(const std::unordered_set<Expression::Kind> kinds);
```
yixinglu, line 20: Why update this?

src/visitor/FindAnyExprVisitor.h
```diff
@@ -68,7 +68,7 @@ class FindAnyExprVisitor final : public ExprVisitorImpl {
 
     bool found_{false};
     const Expression* expr_{nullptr};
-    const std::unordered_set<Expression::Kind>& kinds_;
+    const std::unordered_set<Expression::Kind> kinds_;
```
yixinglu, line 71: ditto

src/visitor/FindAnyExprVisitor.h
jievince: I want use temporary variable:
```
            graph::FindAnyExprVisitor visitor(
                {Expression::Kind::kInputProperty, Expression::Kind::kVarProperty});
```

jievince: Because I want pass a temorary variable to the visitor before:
```
FindAnyExprVisitor visitor({Expression::kind::kInput, Expression::kind::kVarProperty}).
```
But this will introduce an additional copy so I changed it back.

yixinglu (commented):
Could you add some tests in integer vid space like:
```
fetch prop on xx | go from id($-.vertices_)+1 over xx
```

src/parser/Clauses.h
```diff
@@ -92,6 +97,28 @@ class VerticesClause {
...
+                    if (vidList_->size() > 1) {
+                        return false;
+                    }
+                    ref_.reset(vidList_.release()->vidList().back()->clone().release());
+                    DCHECK(vidList_ == nullptr);
+                }
+            }
+        }
+        return true;
+    }
```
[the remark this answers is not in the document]
yixinglu: we only make the tiny function inline.

jievince:
> Could you add some tests in integer vid space like:
> 
> ```
> fetch prop on xx | go from id($-.vertices_)+1 over xx
> ```

It's illegal in parser now:
```
vid
    : unary_integer {
        $$ = new ConstantExpression($1);
    }
    | function_call_expression {
        $$ = $1;
    }
    | uuid_expression {
        $$ = $1;
    }
    | STRING {
        $$ = new ConstantExpression(*$1);
        delete $1;
    }
    ;
vid_ref_expression
    : input_prop_expression {
        $$ = $1;
    }
    | var_prop_expression {
        $$ = $1;
    }
    ;
```

src/parser/Clauses.cpp
```diff
@@ -38,6 +38,37 @@ std::string VertexIDList::toString() const {
...
+        }
+    }
+    return true;
+}
+
+std::string VerticesClause::toString() const {
+    std::string buf;
+    buf.reserve(256);
+    if (isRef()) {
+        buf += ref_->toString();
```
laura-ding, line 65: directly `return ref_->toString();`

src/parser/Clauses.cpp
```diff
@@ -38,6 +38,37 @@ std::string VertexIDList::toString() const {
...
+    } else {
+        buf += vidList_->toString();
```
laura-ding, line 67: ditto

src/parser/Clauses.cpp
jievince: ok, solved

src/parser/Clauses.cpp
jievince: ok solved

Issue #663, opened 2021-01-21
[openCypher Compatibility] Why must WITH used with an alias?


randomJoe211:
![image](https://user-images.githubusercontent.com/69501902/105311211-57eb7d80-5bf8-11eb-8bbb-2b90d1018d3b.png)

openCypher needs no alias while using WITH:

![image](https://user-images.githubusercontent.com/69501902/105311544-6afe4d80-5bf8-11eb-97bd-003ebe401570.png)

czpmango:
The with clause items require alias, whereas Neo4j just provides the default alias.
This is my test:
```
neo4j@neo4j> match (n) with n,1+1 return n;
Expression in WITH must be aliased (use AS) (line 2, column 17 (offset: 17))
"match (n) with n,1+1 return n;"
                 ^
```

czpmango:
Same as this issue: [https://github.com/vesoft-inc/nebula-graph/issues/624](https://github.com/vesoft-inc/nebula-graph/issues/624)

randomJoe211:
Can we have a default alias as well?

czpmango:
> Can we have a default alias as well?

Feature.

randomJoe211:
It's also a compatibility problem that makes the use of openCypher syntax in nGQL troublesome.

czpmango:
> It's also a compatibility problem that makes the use of openCypher syntax in nGQL troublesome.

ACK.

Issue #624, opened 2021-01-15
With clause can not provide a default alias


czpmango:
```

(czp@nebula) [nba]> MATCH p = (a:player {name: "Tim Duncan"})-[:like]->(other:player) WHERE other <> a WITH a, other AS other, length(p) AS len RETURN a.name AS name, avg(other.age) AS others, len 
[ERROR (-12)]: SemanticError: Expression in WITH must be aliased (use AS)

Fri, 15 Jan 2021 16:26:28 CST

(czp@nebula) [nba]> MATCH p = (a:player {name: "Tim Duncan"})-[:like]->(other:player) WHERE other <> a WITH a AS a, other AS other, length(p) AS len RETURN a.name AS name, avg(other.age) AS others, len 
+--------------+--------+-----+
| name         | others | len |
+--------------+--------+-----+
| "Tim Duncan" | 38.5   | 1   |
+--------------+--------+-----+
```

PR #742, opened 2021-02-05
[Bugfix] Label expression in with clause has default alias: its label name

jievince:
fix https://github.com/vesoft-inc/nebula-graph/issues/663, fix https://github.com/vesoft-inc/nebula-graph/issues/624

### DIFF
--- a/src/validator/MatchValidator.cpp
+++ b/src/validator/MatchValidator.cpp
@@ -451,7 +451,11 @@ Status MatchValidator::validateWith(const WithClause *with,
     exprs.reserve(with->columns()->size());
     for (auto *col : with->columns()->columns()) {
         if (col->alias() == nullptr) {
-            return Status::SemanticError("Expression in WITH must be aliased (use AS)");
+            if (col->expr()->kind() == Expression::Kind::kLabel) {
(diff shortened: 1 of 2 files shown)


src/validator/MatchValidator.cpp
```diff
@@ -450,6 +450,9 @@ Status MatchValidator::validateWith(const WithClause *with,
     std::vector<const Expression *> exprs;
     exprs.reserve(with->columns()->size());
     for (auto *col : with->columns()->columns()) {
+        if (col->expr()->kind() == Expression::Kind::kLabel) {
+            col->setAlias(new std::string(col->expr()->toString()));
+        }
         if (col->alias() == nullptr) {
```
Shylock-Hg, line 456: I think you could remove this check.

jievince: Just add an alias for `Label Expression` implicitly. So this check is necessary

yixinglu (approved):

bright-starry-sky (approved):
LGTM

Shylock-Hg (approved):
LGTM

PR #748, opened 2021-02-18
Refactor Iterator

jievince:
The background is in https://confluence.nebula-graph.io/pages/viewpage.action?pageId=8093709
⚠️：The `Iterator` names in current pr have not been modified yet. eg.`Iterator` should be renamed to `Container`,`GetNeighborsIter` should be renamed to `GetNeighborsContainer`...


CPWstatic:
Thanks for the contribution, since we had delete the `LogicalRow` for considering the performance, please redesign the implement of sort and topn.

PR #752, opened 2021-02-19
fix withClause

czpmango:
Fixes [https://jira.nebula-graph.io/browse/NG-435](https://jira.nebula-graph.io/browse/NG-435)

This PR fixes the bug that the order by operator becomes invalid due to unstable filtering in the filter executor.


tests/tck/features/match/WithUnwind.feature
```diff
@@ -186,3 +186,36 @@ Feature: With clause and Unwind clause
...
       | "hello" | 3 | 3 |
+
+  Scenario: match with return
+    When executing query:
+      """
+      MATCH (v:player)
+      WITH v.age AS age, v AS v, v.name AS name
+         ORDER BY age DESCENDING, name ASCENDING
+         LIMIT 20
+         WHERE age > 30
```
Shylock-Hg, line 197: Is there order requirement of order by and where ? I think filter first then sort is equivalent.

czpmango: The where clause in the with clause must be executed last, which is specified by the openCypher standard. In fact, the where clause here is more like the having clause in sql.
You can refer to [https://neo4j.com/docs/cypher-manual/current/clauses/with/#with-filter-on-aggregate-function-results](https://neo4j.com/docs/cypher-manual/current/clauses/with/#with-filter-on-aggregate-function-results) for details.

src/executor/query/FilterExecutor.cpp
```diff
@@ -38,7 +38,11 @@ folly::Future<Status> FilterExecutor::execute() {
                                  "the type should be NULL,EMPTY or BOOL");
         }
         if (val.empty() || val.isNull() || !val.getBool()) {
-            iter->unstableErase();
+            if (UNLIKELY(filter->needStableFilter())) {
```
jievince, line 41: Why use `UNLIKELY`?

czpmango: Just provide dynamic branch prediction hints at compile time.

jievince (approved):

CPWstatic (approved):
LGTM.

PR #753, opened 2021-02-19
Correct the git sha in http server.

Shylock-Hg:
Require https://github.com/vesoft-inc/nebula-common/pull/415

Don't ENABLE_BUILID_STORAGE for the correct git sha that storage build process will overwrite the git info sha of graph.


yixinglu:
we couldn't share the same version file in  common repo since graph and storage will use the common repo git sha which is not our expected result.

Shylock-Hg:
> we couldn't share the same version file in common repo since graph and storage will use the common repo git sha which is not our expected result.

The sha is passed to common from graph/storage.

yixinglu:
> > we couldn't share the same version file in common repo since graph and storage will use the common repo git sha which is not our expected result.
> 
> The sha is passed to common from graph/storage.

ok, LGTM

yixinglu (dismissed):

Aiee (dismissed):

Aiee (approved):

yixinglu (approved):

Issue #707, opened 2021-01-28
client crashed with a query


whitewum:
![image](https://user-images.githubusercontent.com/50101159/106106430-7c0f0780-6180-11eb-82e8-4b6232efc0cb.png)

```
with collect_set([1,2,3]) as a unwind a as b return b
```

My aim is to test whether collect returns a {} or a [] ?

czpmango:
Same as [https://github.com/vesoft-inc/nebula-graph/issues/648](https://github.com/vesoft-inc/nebula-graph/issues/648)

Issue #668, opened 2021-01-21
using count() in WITH causes a crash


randomJoe211:
![image](https://user-images.githubusercontent.com/69501902/105340727-3e0e6280-5c19-11eb-8a52-d9be9635b3d3.png)

czpmango:
Not supported yet. Same as [https://github.com/vesoft-inc/nebula-graph/issues/648](https://github.com/vesoft-inc/nebula-graph/issues/648).

Issue #648, opened 2021-01-20
[Unsupported] with clause groupby


czpmango:
crash.

```
WITH collect([0, 0.0]) AS n return n
```

PR #763, opened 2021-02-23
Add support to use Aggregation functions with WITH clause

Aiee:
As title.

Fix problems caused by incompatibility: 
fix https://github.com/vesoft-inc/nebula-graph/issues/648 
fix https://github.com/vesoft-inc/nebula-graph/issues/707 
fix https://github.com/vesoft-inc/nebula-graph/issues/668

### DIFF
--- a/tests/tck/features/match/WithUnwind.feature
+++ b/tests/tck/features/match/WithUnwind.feature
@@ -37,6 +37,53 @@ Feature: With clause and Unwind clause
       """
     Then a SemanticError should be raised at runtime: Expression in WITH must be aliased (use AS)
 
+  Scenario: with agg return
+    When executing query:
+      """
+      WITH collect([0, 0.0, 100]) AS n
+      RETURN n
+      """
+    Then the result should be, in any order:
+      | n               |
+      | [[0, 0.0, 100]] |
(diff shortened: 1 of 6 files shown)


tests/tck/features/match/WithUnwind.feature
```diff
@@ -37,6 +37,54 @@ Feature: With clause and Unwind clause
       """
     Then a SemanticError should be raised at runtime: Expression in WITH must be aliased (use AS)
 
+  Scenario: with agg return
```
czpmango, line 40: You could add some `nested aggregate function` test cases, such as `with avg(v.age)+1 AS age`.

czpmango (approved):
Well Done!

tests/tck/features/match/WithUnwind.feature
Aiee: The nested situation is included in 
```
      """
      MATCH (v :player{name:"Tim Duncan"})-[]-(v2)
      WITH avg(v2.age) as average_age
      RETURN average_age
      """
```

czpmango (approved):

yixinglu (approved):

PR #764, opened 2021-02-23
Optimize/lifetime

Shylock-Hg:
NG-445

https://jira.nebula-graph.io/browse/NG-445

Reduce about 18% memory usage in a simply match test case, and no effect to latency.


src/context/QueryContext.h
```diff
@@ -124,6 +124,27 @@ class QueryContext {
...
+
+    StatusOr<int64_t> lastUser(const std::string &var) const {
+        auto find = lastUser_.find(var);
+        if (find == lastUser_.end()) {
+            return Status::Error("Can't find variable in user table.");
+        } else {
+            return find->second;
+        }
+    }
+
```
CPWstatic, line 147: I think we could put this part in symtable.

src/context/Result.h
```diff
@@ -45,22 +49,44 @@ class Result final {
...
+            msg = c.msg;
+            values = c.values;
+            iter = c.iter->copy();
+            return *this;
+        }
+
         State state;
         std::string msg;
-        std::shared_ptr<Value> value;
+        std::vector<std::shared_ptr<Value>> values;
```
CPWstatic, line 89: Why we need a vector?

src/executor/query/DedupExecutor.cpp
```diff
@@ -15,31 +15,29 @@ folly::Future<Status> DedupExecutor::execute() {
     SCOPED_TIMER(&execTime_);
     auto* dedup = asNode<Dedup>(node());
     DCHECK(!dedup->inputVar().empty());
-    auto iter = ectx_->getResult(dedup->inputVar()).iter();
+    Result result = ectx_->getResult(dedup->inputVar());
 
-    if (UNLIKELY(iter == nullptr)) {
+    if (UNLIKELY(result.iterRef() == nullptr)) {
```
CPWstatic, line 20: I don't think this is a good practice. Why not use the iterator directly, but call `result.iterRef` for several times?

src/context/Result.h
Shylock-Hg: Result should hold the all value it used, e.g. union sequential iterator will hold two independent value.

src/executor/query/DedupExecutor.cpp
Shylock-Hg: Some result will reference the value like `filter`, so we could copy the origin result and save it.

src/context/QueryContext.h
Shylock-Hg: Fixed.

src/context/Result.h
CPWstatic: I don't see any usage in current pr. Besides, the union does not hold  two value.

Shylock-Hg: Ok, the implementation of union had be modified after this pr. I will check this usage later.

Shylock-Hg: Fixed.

CPWstatic:
close #1053

src/context/Result.h
```diff
@@ -58,6 +66,20 @@ class Result final {
     }
 
     struct Core {
+        Core() = default;
+        Core(Core &&) = default;
+        Core& operator=(Core &&) = default;
+        Core(const Core &c) {
+            *this = c;
+        }
+        Core& operator=(const Core &c) {
```
yixinglu, line 75: check `&c == this` to avoid self assignment.

src/executor/Executor.h
```diff
@@ -20,6 +20,8 @@
 #include "context/ExecutionContext.h"
 #include "util/ScopedTimer.h"
 
+DECLARE_bool(enable_lifetime_optimize);
```
yixinglu, line 23: Need not to declare this flag here

src/context/Symbols.h
```diff
@@ -118,12 +119,36 @@ class SymbolTable final {
         }
     }
 
+    struct VarUser {
+        int64_t id;
+        bool    inLoop;
+    };
```
yixinglu, line 125: Why not place this usage info into Variable structure ?

src/executor/query/DedupExecutor.cpp
yixinglu: I have some thought about the `iterRef` interface. 

firstly, we should consider why we define a `iter` function which always returns its copy. If we must let user to access the inner data, why to return the shared_ptr rather than unique_ptr? It maybe represent that we don't do well designs if we need to provide so many functions such as  `iter` and `iterRef`.

Shylock-Hg: This is only temporary method. I think we just provide method `iter`, and user need manually call `copy` if need a copy.

Shylock-Hg: I don't modify it completely for it's not related to this PR, will fix it later.

src/context/Result.h
Shylock-Hg: Fixed.

src/executor/Executor.h
Shylock-Hg: Fixed.

src/context/Symbols.h
Shylock-Hg: Good catch. Modified.

src/executor/Executor.h
Shylock-Hg: Modified.

CPWstatic (approved):

nevermore3 (approved):

Issue #674, opened 2021-01-22
string filter with null


czpmango:
nebula test:
```
(czp@nebula) [nba]> insert vertex player(name,age) values " \nnull":(null ,0)
Execution succeeded (time spent 29359/31600 us)

(czp@nebula) [nba]> match p = (a:player)-[:like]->(b) where a.name == null return a.name
 Crash!
```

jievince:
```match (v:player) where v.name == null return v```
will cause crash too.

Aiee:
same as https://github.com/vesoft-inc/nebula-graph/issues/721

PR #769, opened 2021-02-24
Fix filter string with null

czpmango:
Fixes [https://jira.nebula-graph.io/browse/NG-428](https://jira.nebula-graph.io/browse/NG-428) and fixes [https://github.com/vesoft-inc/nebula-graph/issues/674](https://github.com/vesoft-inc/nebula-graph/issues/674)


src/optimizer/OptimizerUtils.cpp
```diff
@@ -445,6 +445,9 @@ Value OptimizerUtils::normalizeValue(const meta::cpp2::ColumnDef& col, const Val
                 col.get_type().get_type_length() == nullptr) {
                 return Value::kNullBadType;
             }
+            if (!v.isStr()) {
+                return Value::kNullValue;
```
czpmango, line 449: `return v` or `return null` ?

jievince:
`MATCH (v:player) where v.name == nullptr` will cause graphd to crash is because this statement will generate a `IndexScan` node. However, the filter of `IndexScan` is strict. So I think if the filter is not satisfied with `IndexScan`, it should not generate a `IndexScan`

src/optimizer/OptimizerUtils.cpp
```diff
@@ -445,6 +445,9 @@ Value OptimizerUtils::normalizeValue(const meta::cpp2::ColumnDef& col, const Val
                 col.get_type().get_type_length() == nullptr) {
                 return Value::kNullBadType;
             }
+            if (!v.isStr()) {
+                return v;
```
jievince, line 449: return Value::kNullBadType?

czpmango: > return Value::kNullBadType?

This function is just to normalize the value in the opt phase.For relational expressions, comparison of different types is also a normal input scenario.

yixinglu (approved):

bright-starry-sky:
Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
this issue same with : 
https://github.com/vesoft-inc/nebula-graph/pull/777

A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.

czpmango:
> Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
> this issue same with :
> #777
> 
> A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.



> Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
> this issue same with :
> #777
> 
> A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.



> Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
> this issue same with :
> #777
> 
> A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.

This PR is just to fix crash bug, which caused by illegal use of the `getStr()` interface when the value is not a string.
So it has nothing to do with `lookupValidator::checkConstExpr`.

Aiee:
> > Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
> > this issue same with :
> > #777
> > A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.
> 
> This PR is just to fix crash bug, which caused by illegal use of the `getStr()` interface when the value is not a string.
> So it has nothing to do with `lookupValidator::checkConstExpr`.

I believe what @bright-starry-sky saying is that should check the type in the validation stage and do typecasting in the validator so that optimizer only receives data that is pre-processed.

czpmango:
> > > Actually, This bug caused by `LookupValidator::checkConstExpr`. I don't know why it's not blocked at here.
> > > this issue same with :
> > > #777
> > > A suggestion, Improve the method `LookupValidator::checkConstExpr`, let it block the invalid data type.
> > 
> > 
> > This PR is just to fix crash bug, which caused by illegal use of the `getStr()` interface when the value is not a string.
> > So it has nothing to do with `lookupValidator::checkConstExpr`.
> 
> I believe what @bright-starry-sky saying is that we should check the type in the validation stage and do typecasting in the validator so that optimizer only receives data that is pre-processed.

That is another logic. The function `OptimizerUtils::normalizeValue` is just for normalization of const value rather than typecasting.

bright-starry-sky:
After offline discussion with @Aiee about #777, I think the both are similar logic.You can discuss it with @Aiee , avoid conflict at work.
I think , if you want to fix the crash problem, you just need to check OptimizerUtils::normalizeValue in validator.
if you want to support null type, you need to implement the code with data type null.
In this PR, simply passing this type now will lead to other problems later.

jievince (approved):

Issue #762, opened 2021-02-23
update/upsert when statement improves


panda-sheep:
```
create space y2(partition_num=1, replica_factor=1, vid_type=int64);
use y2
create tag tag2(col1 int,col2 bool,col3 string);
insert vertex tag2(col1,col2,col3) values 1:(1,true,"aaa");

(user@nebula) [y2]> fetch prop on tag2 1
+----------+-----------+-----------+-----------+
| VertexID | tag2.col1 | tag2.col2 | tag2.col3 |
+----------+-----------+-----------+-----------+
| 1        | 1         | true      | "aaa"     |
+----------+-----------+-----------+-----------+
Got 1 rows (time spent 3783/4221 us)

update vertex on tag2 1 set col3="bbb" when col1;  
```

`update vertex on tag2 1 set col3="bbb" when col1;`
This statement should report a syntax error

laura-ding:
> ```
> create space y2(partition_num=1, replica_factor=1, vid_type=int64);
> use y2
> create tag tag2(col1 int,col2 bool,col3 string);
> insert vertex tag2(col1,col2,col3) values 1:(1,true,"aaa");
> 
> (user@nebula) [y2]> fetch prop on tag2 1
> +----------+-----------+-----------+-----------+
> | VertexID | tag2.col1 | tag2.col2 | tag2.col3 |
> +----------+-----------+-----------+-----------+
> | 1        | 1         | true      | "aaa"     |
> +----------+-----------+-----------+-----------+
> Got 1 rows (time spent 3783/4221 us)
> 
> update vertex on tag2 1 set col3="bbb" when col1;  
> ```
> 
> `update vertex on tag2 1 set col3="bbb" when col1;`
> This statement should report a syntax error

why? `when col1` is right expression.

panda-sheep:
https://github.com/vesoft-inc/nebula-common/pull/412 
This pr requires `when/where` must be followed by：
1) bool expression
2) Bool type column
3) String type column

czpmango:
> [vesoft-inc/nebula-common#412](https://github.com/vesoft-inc/nebula-common/pull/412)
> This pr requires `when/where` must be followed by：
> 
> 1. bool expression
> 2. Bool type column
> 3. String type column

This issue has nothing to do with pr 412. Should follow update/upsert’s own implementation logic, and I don’t think the implementation of update/upsert needs to call type casting.

czpmango:
> > ```
[...]

PR #772, opened 2021-02-25
fix update/upsert

czpmango:
Add type check for when clause in update/upsert statement.
Expression in when clause should return bool.

Fixes [https://github.com/vesoft-inc/nebula-graph/issues/762](https://github.com/vesoft-inc/nebula-graph/issues/762)

### DIFF
--- a/src/validator/test/MutateValidatorTest.cpp
+++ b/src/validator/test/MutateValidatorTest.cpp
@@ -164,7 +164,7 @@ TEST_F(MutateValidatorTest, UpdateVertexTest) {
     {
         auto cmd = "UPDATE VERTEX ON person \"Tom\""
                    "SET age = age + 1 "
-                   "WHEN page == 18 "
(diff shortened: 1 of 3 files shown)


jievince (approved):
LGTM

src/validator/test/MutateValidatorTest.cpp
```diff
@@ -164,7 +164,7 @@ TEST_F(MutateValidatorTest, UpdateVertexTest) {
     {
         auto cmd = "UPDATE VERTEX ON person \"Tom\""
                    "SET age = age + 1 "
-                   "WHEN page == 18 "
```
Shylock-Hg, line 167: This don't report error?

czpmango: The previous implementation will not report an error.
I added the deduceType check after rewrited this Expression.

yixinglu (approved):

Issue #721, opened 2021-02-01
Type check failed in match where clause


czpmango:
```
(czp@nebula) [nba]>       MATCH p = (n:player)<-[e:like]-(m)       WHERE n.age >= 33.0000000000011         RETURN n.name AS player, n.age AS age
[ERROR (-12)]: SemanticError: Data type error of field : age

Mon, 01 Feb 2021 11:06:47 CST

(czp@nebula) [nba]>       MATCH p = (n:player)<-[e:like]-(m)       WHERE n.age >= 33         RETURN n.name AS player, n.age AS age
+---------------------+-----+
| player              | age |
+---------------------+-----+
| "Shaquile O'Neal"   | 47  |
+---------------------+-----+
| "Kobe Bryant"       | 40  |
+---------------------+-----+
| "Kobe Bryant"       | 40  |
+---------------------+-----+

```

jievince:
The filter condition of `IndexScan` is too strict.

Aiee:
https://github.com/vesoft-inc/nebula-graph/pull/784 only fixed LOOKUP. This issue in MATCH retains.

HarrisChu:
MATCH bug still in v2.0.1

PR #777, opened 2021-03-02
Fix indexScan rule when compare INT with Double

Aiee:
1. Fix crash caused by comparing different numerical types in the index scan. Issue: fixes https://github.com/vesoft-inc/nebula-graph/issues/721

TODO:
Fix edges cases failure due to the float precision.

PR #784, opened 2021-03-04
Add support to compare different numerical types in LOOKUP WHERE clause

Aiee:
1. Refactor the implementation of https://github.com/vesoft-inc/nebula-graph/pull/777 and fix https://github.com/vesoft-inc/nebula-graph/issues/721
2. Add support to LOOKUP queries like ```LOOKUP ON player WHERE player.age >= 40.5 YIELD player.age AS Age```
3. The main idea is to check/cast different numerical values in the validation stage so that optimizer and indexScan rule can keep the current logic.

TODOs:
1. Add the same capability to ```MATCH```
2. extract the typecast logic as a method if we decide to support more cross-type comparisons.


src/validator/LookupValidator.cpp
```diff
@@ -397,6 +402,24 @@ StatusOr<Value> LookupValidator::checkConstExpr(Expression* expr, const std::str
...
+            auto castVal = v.toInt().first;
+            // edge case col1 >= 40.0, no need to round up
+            if (castVal == v.getFloat()) {
+                return castVal;
+            }
+            return castVal + 1;
+        }
+        return v.toInt().first;
+    }
+
```
bright-starry-sky, line 422: A small suggestion, It's better to extract a method, such as : 
Value castToTargetType(const Value& v, PropertyType targetType) {
swith (v.getType()) {
case int, float {
 return cast v from current type to target type if needed.
}
case ….. {
return bad value;
}
}
}

src/validator/LookupValidator.h
```diff
@@ -44,7 +44,10 @@ class LookupValidator final : public Validator {
 
     Status rewriteRelExpr(RelationalExpression* expr);
 
-    StatusOr<Value> checkConstExpr(Expression* expr, const std::string& prop);
+    StatusOr<Value> checkConstExpr(Expression* expr,
+                                   const std::string& prop,
+                                   const Expression::Kind kind,
+                                   const bool leftIsAE);
```
bright-starry-sky, line 50: Please think about it. Is it necessary to change it here?

bright-starry-sky:
Good job.

src/validator/LookupValidator.h
Aiee: To determine whether a value should be cast, it is necessary to get the info about the type of relational expression and also the position of the label attribute.

bright-starry-sky: > To determine whether a value should be cast, it is necessary to get the info about the type of relational expression and also the position of the label attribute.

I think auto `type = schema->getFieldType(prop); `is enough. WDYT?

Aiee: ```schema->getFieldType(prop)``` returns the data type of the column while the parameter```kind```is the kind of relational expression(GT, LT). In this function, ```Expression* expr``` is always a const expression, and I couldn't see another way to determine the comparison operator. Please correct me if I am wrong.

src/validator/LookupValidator.cpp
Aiee: The reason I didn't extract a function is that the only case we allow different types to compare is when both are numerical, specifically, int and float. I'll modify this part if we decide to support more cross-type comparisons. 
Does it make sense to you?

src/validator/LookupValidator.cpp
```diff
@@ -397,6 +402,27 @@ StatusOr<Value> LookupValidator::checkConstExpr(Expression* expr, const std::str
...
+    } else if (graph::SchemaUtil::propTypeToValueType(type) == Value::Type::INT && v.isFloat()) {
+        // col1 < 10.5 range: [min, 11), col1 < 10 range: [min, 10)
+        // cast float to int and add 1 to get the ceilling
+        if ((leftIsAE && (kind == Expression::Kind::kRelGE || kind == Expression::Kind::kRelLT)) ||
+            (!leftIsAE && (kind == Expression::Kind::kRelGT || kind == Expression::Kind::kRelLE))) {
+            auto castVal = v.toInt();
+            // edge case col1 >= 40.0, no need to round up
+            if (castVal == v.getFloat()) {
+                return castVal;
+            }
```
yixinglu, line 420:
```cpp
auto f = v.getFloat();
auto i = std::ceil(f);
if (std::abs(f - i) < epsilon) {
  // xxx
}
// yyy
```

src/validator/LookupValidator.cpp
```diff
@@ -387,7 +389,10 @@ Status LookupValidator::rewriteRelExpr(RelationalExpression* expr) {
     return Status::OK();
 }
 
-StatusOr<Value> LookupValidator::checkConstExpr(Expression* expr, const std::string& prop) {
+StatusOr<Value> LookupValidator::checkConstExpr(Expression* expr,
+                                                const std::string& prop,
+                                                const Expression::Kind kind,
+                                                const bool leftIsAE) {
```
yixinglu, line 395: remove `const` qualifier for primitive parameter.

bright-starry-sky (approved):
LGTM now， let ours handle other type later.

yixinglu (approved):

PR #785, opened 2021-03-05
disable_toss

liuyu85cn:
for 2.0 GA, disable the user option to use toss .


src/validator/AdminValidator.cpp
```diff
@@ -84,7 +84,8 @@ Status CreateSpaceValidator::validateImpl() {
             }
             case SpaceOptItem::ATOMIC_EDGE: {
                 if (item->getAtomicEdge()) {
-                    spaceDesc_.set_isolation_level(meta::cpp2::IsolationLevel::TOSS);
+                    // for 2.0 GA, no matter how this option set, don't use toss.
+                    spaceDesc_.set_isolation_level(meta::cpp2::IsolationLevel::DEFAULT);
```
bright-starry-sky, line 88: A small question, is it better to return an error directly?

liuyu85cn: good idea

bright-starry-sky (approved):
LGTM

critical27 (approved):

PR #800, opened 2021-03-10
Introduce weak dependency for plan node

yixinglu:
Scheduler will use the weak dependency to solve the following plan execution issue:

```
A -----> B -----> C
  \               ^
   \______________|
```

A depends on B and C, B depends on C, And we should not execute C twice.


yixinglu:
I need to rethink about the scheduler to solve the join dependency issue.

PR #801, opened 2021-03-11
Fix get src prop dead loop in GetNeighborsIter.

CPWstatic:


yixinglu (approved):

Aiee (approved):

Aiee:
Could you also add some cases that match the scenario? 
Such as ```GO FROM "Boris Diaw" OVER serve YIELD $^.player.name```

yixinglu:
> Could you also add some cases that match the scenario?
> Such as `GO FROM "Boris Diaw" OVER serve YIELD $^.player.name`

Could you add this test case to our CI and close that issue?

PR #804, opened 2021-03-11
build docker image of nebula-tools

jievince:
Based on vesoft-inc/nebula-storage#389


yixinglu (commented):
it's better to build this image in nebula storage

yixinglu (approved):

czpmango (approved):

PR #805, opened 2021-03-11
[DO NOT MERGE] fix the case sensitivity of function names

czpmango:
As title.
Depends on [https://github.com/vesoft-inc/nebula-common/pull/442](https://github.com/vesoft-inc/nebula-common/pull/442)


jievince (approved):

laura-ding (approved):

czpmango:
I redefine the map comparator to solve this problem more elegantly, so this PR will be closed.

PR #810, opened 2021-03-15
Push down some expressions to the storage layer

jievince:
Associated with these two PR:
https://github.com/vesoft-inc/nebula-storage/pull/386
https://github.com/vesoft-inc/nebula-common/pull/448


bright-starry-sky:
Please add test cases for lookup and match.

src/visitor/ExtractFilterExprVisitor.h
```diff
@@ -50,6 +50,7 @@ class ExtractFilterExprVisitor final : public ExprVisitorImpl {
 
     bool canBePushed_{true};
     std::unique_ptr<Expression> remainedExpr_;
+    std::string *innerVar_;  // used by predicate, reduce and list comprehension
```
yixinglu, line 53: Where to use this variable?

yixinglu (commented):
Add some tests which could not be pushed down storage

src/visitor/ExtractFilterExprVisitor.h
jievince: sorry, it's unused and should be deleted

jievince:
> Add some tests which could not be pushed down storage

I think if the filter can't be pushed down, the `filter` node and `GN` node  will not be next to each other in the generated execution plan

jievince:
> Please add test cases for lookup and match.

But `IndexScan` node doesn't accept a filter which has a list comprehension expression
```
(root@nebula) [nba]> lookup on player where  player.age in [v in [5,6,7,8] | v + 90]
[ERROR (-12)]: SemanticError: Expression (player.age IN [v IN [5,6,7,8] | (v+90)]) not supported yet
```

tests/tck/features/optimizer/PushFilterDownGetNbrsRule.feature
```diff
@@ -62,6 +62,35 @@ Feature: Push Filter down GetNeighbors rule
       | Project      | 1            |                                 |
       | GetNeighbors | 2            | filter: (serve.start_year>2005) |
       | Start        |              |                                 |
+    When profiling query:
```
czpmango, line 65: Could you add some test cases like this:
```
$var=go from "Tim Duncan" over like yield like._dst as dst,$$.player.age as age;
go from $var.dst over like where like.likeness IN [v IN [1,3] WHERE v+$var.age > 0
```

src/visitor/ExtractFilterExprVisitor.cpp
```diff
@@ -21,8 +21,12 @@ void ExtractFilterExprVisitor::visit(UUIDExpression *) {
...
 }
 
-void ExtractFilterExprVisitor::visit(VariableExpression *) {
-    canBePushed_ = false;
+void ExtractFilterExprVisitor::visit(VariableExpression *expr) {
+    if (static_cast<VariableExpression *>(expr)->isInner()) {
+        canBePushed_ = true;
+    } else {
+        canBePushed_ = false;
+    }
```
yixinglu, line 29:
```cpp
canBePushed_ = expr->isInner();
```

yixinglu (commented):
Is it better to name the variable as `internal`, `local` or `temporary` than `inner`?

tests/tck/features/optimizer/PushFilterDownGetNbrsRule.feature
jievince: But how to profile a query with a `$var`?

src/visitor/ExtractFilterExprVisitor.cpp
jievince: ok

czpmango (approved):

PR #820, opened 2021-03-16
Improve join when single column.

CPWstatic:


dutor (approved):

yixinglu (commented):
I want to know what's difference about performance improvement if only optimize the `hash` function of `List` type, like:

```cpp
int hash<List>::operator()(const List& l) const {
  if (l.values.size() == 1) return std::hash<Value>()(l.values[0]);
  int seed = 0;
  for (auto& v : l.values) {
     seed ^= hash<Value>()(v) + ....;
  }
  return seed;
}
```

nevermore3 (approved):

PR #837, opened 2021-03-18
Refactor plan node dependencies usage

yixinglu:
as title


src/executor/Executor.cpp
```diff
@@ -113,40 +113,23 @@ Executor *Executor::makeExecutor(const PlanNode *node,
...
+        selectExecutor->setThenBody(thenBody);
+        selectExecutor->setElseBody(elseBody);
+    } else if (node->kind() == PlanNode::Kind::kLoop) {
+        auto loop = asNode<Loop>(node);
+        auto body = makeExecutor(loop->body(), qctx, visited);
+        auto loopExecutor = static_cast<LoopExecutor *>(exec);
+        loopExecutor->setLoopBody(body);
+    }
+
+    for (size_t i = 0; i < node->numDeps(); ++i) {
```
Shylock-Hg, line 131: I think the iterator over container is better than the index.

yixinglu: Plan node already have `dep()` function, needs not to provide a `dependencies()` to use.

Aiee (approved):

jievince (approved):

PR #842, opened 2021-03-18
add optimizer rule for having clause

czpmango:
> 

This PR adds an optRule called `PushFilterDownAggregateRule` to optimize the performance of the query statements with `having` clause.

Scenarios for this OptRule:
   Old Plan: A plan that looks like `Start->...->Aggregate->Filter(without aggregate item)->...->Root`
   Plan after applying rule: `Start->...->Filter(without aggregate item)->Aggregate->...->Root`

Example:
```
   match (v:player) with v.age+1 as age, count(v.age) as count where age<3 return age,count
```

---

Performance comparison:
Query: `profile match (v:player) with v.name as name,v.age+1 as age,count(*) as count where age<5 and name>"A" return name,age,count`
Old plan:

Plan after applying rule:

---

TODO: 
  - Supports optimization of statements with nested aggregate functions
  - Support partial push-down of filter


src/optimizer/OptRule.h
```diff
@@ -59,12 +59,12 @@ class OptRule {
         std::vector<OptGroupNode *> newGroupNodes;
     };
 
-    StatusOr<MatchedResult> match(OptContext *ctx, const OptGroupNode *groupNode) const;
+    StatusOr<MatchedResult> match(OptContext *ctx, const OptGroupNode *groupNode);
```
yixinglu, line 62: Why do you need to change these function signatures ? One thing which you should know is that all opt rules should be stateless since they are shared by all queries.

src/optimizer/rule/PushFilterDownAggregateRule.h
```diff
@@ -0,0 +1,39 @@
...
+
+    bool match(OptContext *ctx, const MatchedResult &matched);
+
+private:
+    PushFilterDownAggregateRule();
+
+    static std::unique_ptr<OptRule> kInstance;
+
+    // This map generated in `OptRule::match` will be used in the subsequent `OptRule::transform`
+    std::unordered_map<std::string, Expression *> rewriteMap_;
```
yixinglu, line 33: please rethink what i said above.

src/optimizer/OptRule.cpp
```diff
@@ -71,7 +71,7 @@ StatusOr<MatchedResult> OptRule::match(OptContext *ctx, const OptGroupNode *grou
     return matched;
 }
 
-bool OptRule::match(OptContext *ctx, const MatchedResult &matched) const {
+bool OptRule::match(OptContext *ctx, const MatchedResult &matched) {
```
czpmango, line 74: I removed the const feature of `OptRule::match` bcz I expected the member functions within class `OptRule` could communicate through member variables. This is usually useful in certain optimization scenarios involving expression rewriting.

src/optimizer/OptRule.h
czpmango: How about removing its state after `OptRule::transfrom`?

src/visitor/RewriteAllSubExprVisitor.h
```diff
@@ -0,0 +1,79 @@
...
+#ifndef VISITOR_REWRITEALLSUBEXPRVISITOR_H_
+#define VISITOR_REWRITEALLSUBEXPRVISITOR_H_
+
+#include "common/expression/Expression.h"
+#include "visitor/ExprVisitorImpl.h"
+
+namespace nebula {
+namespace graph {
+
+class RewriteAllSubExprVisitor final : public ExprVisitorImpl {
```
Shylock-Hg, line 16: Plz add a comment to explain the visitor's function.

src/visitor/RewriteAllSubExprVisitor.h
```diff
@@ -0,0 +1,79 @@
...
+public:
+    explicit RewriteAllSubExprVisitor(const Expression* rewrittenSubExpr,
```
czpmango, line 18: format.

src/visitor/RewriteAllSubExprVisitor.h
czpmango: Which function?

src/visitor/RewriteAllSubExprVisitor.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "visitor/RewriteAllSubExprVisitor.h"
+
+namespace nebula {
+namespace graph {
+
+RewriteAllSubExprVisitor::RewriteAllSubExprVisitor(
+                                   const Expression* rewrittenSubExpr,
```
czpmango, line 13: format.

src/optimizer/OptRule.h
czpmango: I missed considering this.  It will really cause some runtime error. I will reimplement it.

czpmango: Already reverted.

src/optimizer/rule/PushFilterDownAggregateRule.h
czpmango: Reverted.

src/visitor/RewriteAllSubExprVisitor.h
Shylock-Hg: Emm.Purpose of the visitor.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // Rewrite VariablePropertyExpr in filter's condition
+    for (const auto& item : rewriteMap_) {
+        const VariablePropertyExpression rewrittenSubExpr(new std::string(),
+                                                          new std::string(item.first));
+        auto* rewriteSubExpr = item.second;
+        graph::RewriteAllSubExprVisitor
+               rewriteConditionVisitor(&rewrittenSubExpr, rewriteSubExpr, condition);
+        condition->accept(&rewriteConditionVisitor);
+        if (rewriteConditionVisitor.needRewriteRoot()) {
+            condition = octx->objPool()->add(rewriteSubExpr->clone().release());
```
yixinglu, line 94: Where to use this condition expression? Note that you should not use objpool of OptContext to manage the memory of plan node because of OptContext will be released after optimization, switch it to QueryContext object pool.

src/visitor/RewriteAllSubExprVisitor.h
```diff
@@ -0,0 +1,80 @@
...
+                                      const Expression* rewriteSubExpr,
+                                      Expression* originExpr);
+
```
yixinglu, line 21: Prefer a `RewriteVarExprVisitor` to replace `$var` expression with _new expression_. Do you check wether these existing RewriteVisitors meet your requirements, such as `RewriteLabelAttrVisitor` or `RewriteSymExprVisitor` ?

yixinglu (commented):
Need some plan test cases, plz check the optimizer folder in features directory

src/visitor/RewriteAllSubExprVisitor.h
czpmango: I have considered the existing visitor, none of them can meet the needs here. I hope to implement a more general visitor, which will rewrite all internal expressions equal to `rewrittenSubExpr` in a certain parrent expression(originExpr here) to `rewriteSubExpr`, and It's not just about judging the type of expression. Furthermore, VariablePropertyExpr here is just a special case, and this is a type-independent visitor.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
czpmango: You are right, my mistake. Already fixed.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,125 @@
...
+    std::unordered_map<std::string, Expression *> rewriteMap_;
+    bool noAggFilter = true;
+    auto colNames = newAggNode->colNames();
+    for (size_t i = 0; i < colNames.size(); ++i) {
+        auto& colName = colNames[i];
+        for (size_t j = 0; j < propNames.size(); ++j) {
+            if (!colName.compare(propNames[j])) {
+                if (graph::ExpressionUtils::findAny(groupItems[i],
+                                                   {Expression::Kind::kAggregate})) {
+                    noAggFilter = false;
```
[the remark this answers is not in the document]
yixinglu: And I feel you can simplify this loop logic by `find` to avoid to reassignment of `rewriteMap_[colName]` since the inner loop is not related to the iterable variable `j`. such as:

```cpp
auto colName = colNames[i];
auto found = std::find_if(propNames.begin(), propNames.end(), [&colName](auto &prop) { return !colName.compare(prop); });
if (found != propNames.end()) {
  if (graph::...::findAny(xxx)) {
    return noTransformResult();
  }
  rewriteMap_[colName] = groupItems[i];
}

```

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // Rewrite VariablePropertyExpr in filter's condition
+    for (const auto& item : rewriteMap_) {
+        const VariablePropertyExpression rewrittenSubExpr(new std::string(),
+                                                          new std::string(item.first));
+        auto* rewriteSubExpr = item.second;
+        graph::RewriteAllSubExprVisitor
+               rewriteConditionVisitor(&rewrittenSubExpr, rewriteSubExpr, condition);
+        condition->accept(&rewriteConditionVisitor);
+        if (rewriteConditionVisitor.needRewriteRoot()) {
+            condition = qctx->objPool()->add(rewriteSubExpr->clone().release());
```
yixinglu, line 94: I still don't find where the `condition` is used in the following parts of the `transform` function. So why to assign to it here ?

src/visitor/RewriteAllSubExprVisitor.cpp
```diff
@@ -0,0 +1,121 @@
...
+    DCHECK(!!rewrittenSubExpr);
+    DCHECK(!!rewriteSubExpr);
+    DCHECK(!!originExpr);
+    needRewriteRoot_ = *rewrittenSubExpr_ == *originExpr_;
+}
+
+void RewriteAllSubExprVisitor::visit(TypeCastingExpression *expr) {
+    if (needRewriteRoot_) return;
+    auto* operand = expr->operand();
+    if (*operand == *rewrittenSubExpr_) {
```
yixinglu, line 28: It's really confused to compare two expression objects directly? It's dangerous for current expression implementation which use many pointers in it if one expression doesn't overwrite the `operator==`. I feel you should rethink how to rewrite or replace some subexpression by an existing expression.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
czpmango: `condition` is in newFilterNode.

src/visitor/RewriteAllSubExprVisitor.h
czpmango: I plan to abstract and unify all rewrite visitors, and the visitor here can also be deleted after completion.
Refer to [PR#870](https://github.com/vesoft-inc/nebula-graph/pull/870) for details about visitor refactoring.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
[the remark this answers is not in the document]
czpmango: ACK.

src/visitor/RewriteAllSubExprVisitor.cpp
czpmango: ACK. I implemented a more abstract visitor in  [PR#870](https://github.com/vesoft-inc/nebula-graph/pull/870) , so the visitor here will be abandoned.

src/visitor/RewriteAllSubExprVisitor.h
czpmango: Done.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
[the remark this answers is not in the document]
czpmango: Done.

src/visitor/RewriteAllSubExprVisitor.cpp
czpmango: Already delete this visitor.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,127 @@
...
+    auto* filterGroupNode = matched.node;
+    auto* oldFilterNode = filterGroupNode->node();
+    auto deps = matched.dependencies;
+    DCHECK_EQ(deps.size(), 1);
+    auto aggGroupNode = deps.front().node;
+    auto* oldAggNode = aggGroupNode->node();
+    DCHECK(oldFilterNode->kind() == PlanNode::Kind::kFilter);
+    DCHECK(oldAggNode->kind() == PlanNode::Kind::kAggregate);
+    auto* newFilterNode =
+        const_cast<graph::Filter*>(static_cast<const graph::Filter*>(oldFilterNode))->clone(qctx);
```
yixinglu, line 49: Need not to cast the pointer to nonconst Filter since clone is a const function

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,127 @@
...
+    auto* newAggNode =
+        const_cast<graph::Aggregate*>(static_cast<const graph::Aggregate*>(oldAggNode))
```
yixinglu, line 51: ditto

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,127 @@
...
+    auto colNames = newAggNode->colNames();
+    for (size_t i = 0; i < colNames.size(); ++i) {
+        auto& colName = colNames[i];
+        for (size_t j = 0; j < propNames.size(); ++j) {
+            if (!colName.compare(propNames[j])) {
+                if (graph::ExpressionUtils::findAny(groupItems[i],
+                                                    {Expression::Kind::kAggregate})) {
+                    return TransformResult::noTransform();
+                } else {
+                    rewriteMap[colName] = groupItems[i];
```
yixinglu, line 73: break;

yixinglu: plz rethink my former suggestion. I really advise you to refactor the logics with `std::find_if` such as:

```cpp
    for (size_t i = 0; i < colNames.size(); ++i) {
        auto& colName = colNames[i];
        auto iter = std::find_if(propNames.begin(), propNames.end(), [&](const auto& name) {
            return !colName.compare(name);
        });
        if (iter == propNames.end()) continue;
        if (graph::ExpressionUtils::findAny(groupItems[i], {Expression::Kind::kAggregate})) {
            return TransformResult::noTransform();
        }
        rewriteMap[colName] = groupItems[i];
    }
```

It's more clear without many nested blocks.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // Rewrite VariablePropertyExpr in filter's condition
+    auto matcher = [&rewriteMap](const Expression* e) -> bool {
+        if (e->kind() != Expression::Kind::kVarProperty) {
+            return false;
+        }
+        auto* propName = static_cast<const VariablePropertyExpression*>(e)->prop();
+        if (!rewriteMap[*propName]) {
+            return false;
+        }
+        return true;
```
yixinglu, line 88: `return rewriteMap[*propName]` directly

src/optimizer/rule/PushFilterDownAggregateRule.h
```diff
@@ -0,0 +1,34 @@
...
+#include "optimizer/OptRule.h"
+
+namespace nebula {
+namespace opt {
+
+class PushFilterDownAggregateRule final : public OptRule {
+public:
+    const Pattern &pattern() const override;
+
+    StatusOr<OptRule::TransformResult> transform(OptContext *         qctx,
```
yixinglu, line 20: format this file.

src/planner/Query.cpp
```diff
@@ -93,7 +92,7 @@ std::unique_ptr<PlanNodeDescription> GetVertices::explain() const {
 }
 
 GetVertices* GetVertices::clone() const {
-    auto src = qctx_->objPool()->add(src_->clone().release());
+    auto                                   src = qctx_->objPool()->add(src_->clone().release());
```
yixinglu, line 95: ditto

src/planner/Query.cpp
```diff
@@ -271,13 +299,13 @@ std::unique_ptr<PlanNodeDescription> DataCollect::explain() const {
     return desc;
 }
 
-Join::Join(QueryContext* qctx,
-           Kind kind,
-           PlanNode* input,
+Join::Join(QueryContext*                   qctx,
```
yixinglu, line 302: plz revert this format

src/planner/Query.h
```diff
@@ -71,13 +71,13 @@ class Explore : public SingleInputNode {
...
 
 protected:
-    Explore(QueryContext* qctx,
-            Kind kind,
-            PlanNode* input,
-            GraphSpaceID space,
-            bool dedup,
-            int64_t limit,
-            std::string filter,
+    Explore(QueryContext*                       qctx,
```
yixinglu, line 74: and all this whole file.

yixinglu (commented):
Remember that only format what you changed

src/optimizer/rule/PushFilterDownAggregateRule.cpp
czpmango: ok

src/optimizer/rule/PushFilterDownAggregateRule.cpp
czpmango: It is just a matcher of `RewriteVisitor`.

czpmango: ACK.

src/optimizer/rule/PushFilterDownAggregateRule.cpp
czpmango: Modified.

yixinglu (dismissed):

yixinglu (approved):

Aiee (approved):

PR #843, opened 2021-03-18
Support group name when create space

darionyaphet:


jievince (approved):

yixinglu (dismissed):

src/validator/AdminValidator.cpp
```diff
@@ -22,6 +22,9 @@ Status CreateSpaceValidator::validateImpl() {
     ifNotExist_ = sentence->isIfNotExist();
     auto status = Status::OK();
     spaceDesc_.space_name = std::move(*(sentence->spaceName()));
+    if (sentence->groupName() != nullptr) {
+        spaceDesc_.set_group_name(std::move(*(sentence->groupName())));
+    }
```
bright-starry-sky, line 27: Do we need to check the group exists?

darionyaphet: sure it is an optional

PR #855, opened 2021-03-22
Avoid build storage for graphd image building. And try the local modu…

Shylock-Hg:
…les first.

### DIFF
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -18,6 +18,8 @@ cmake_minimum_required(VERSION 3.5.0)
 # Set the project name
 project("Nebula Package")
 
+OPTION(ENABLE_PACKAGE_STORAGE "Whether package storage." ON)
+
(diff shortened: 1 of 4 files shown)


package/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ cmake_minimum_required(VERSION 3.5.0)
 # Set the project name
 project("Nebula Package")
 
+OPTION(ENABLE_PACKAGE_STORAGE "Whether package storage." ON)
```
yixinglu, line 21: Why not use `ENABLE_BUILD_STORAGE` to control the package process ?

Shylock-Hg: It's not the option of this camke file.

yixinglu: Could the CMakeList.txt in subdirectory not use global option ?

Shylock-Hg: They are different cmake building.This is just for package.

yixinglu (approved):

laura-ding (approved):

PR #867, opened 2021-03-23
Support MATCH using partial indexes in filter

Aiee:
As title. If there are multiple indexes in the where clause and no composite index has been built, the left-most index will be used for the query.

Need to discuss whether this change is appropriate: https://docs.nebula-graph.io/2.0/3.ngql-guide/7.general-query-statements/5.lookup/
```
No valid index found¶

No valid index found
If your query filter contains a string type field, Nebula Graph selects the index that matches all the fields. For example:


nebula> CREATE TAG t1 (c1 string, c2 int);
nebula> CREATE TAG INDEX i1 ON t1 (c1, c2);
nebula> LOOKUP ON t1 WHERE t1.c1 == "a"; -- Index i1 is invalid
nebula> LOOKUP ON t1 WHERE t1.c1 == "a" and t1.c2 == 1;  -- Index i1 is valid
```
e.g.  
MATCH
```
(root@nebula) [nba]> MATCH (v:player) WHERE v.age>40 AND v.name == "Tim Duncan" RETURN v.name, v.age

+--------------+-------+
| v.name       | v.age |
+--------------+-------+
| "Tim Duncan" | 42    |
+--------------+-------+
Got 1 rows (time spent 6305/6716 us)
```

LOOKUP
```
(root@nebula) [nba]> profile lookup on player where player.name == "Tim Duncan" and player.age>40
+--------------+
| VertexID     |
+--------------+
| "Tim Duncan" |
+--------------+
Got 1 rows (time spent 4219/4690 us)

Execution Plan (optimize time 187 us)
```

TODO:
- [ ] Support to use any valid index rather than the left-most one only.


Aiee:
Need to discuss the approach.

PR #873, opened 2021-03-25
Generate log directory, if it does not exist

panda-sheep:
As title

When we start the service with the following script, the log directory will be created in the script.
./scripts/nebula.service -c xxx/etc/nebula-graphd.conf start graphd> /dev/null 2>&1 &

But when we start the service in other ways, because the directory log does not exist, the startup fails.
xxx/bin/nebula-graphd --flagfile xxx/etc/nebula-graphd-1.conf >1.log 2 >&1 &

Therefore, before starting the service, the log directory needs to be created automatically when it does not exist.

### DIFF
--- a/conf/nebula-graphd.conf.default
+++ b/conf/nebula-graphd.conf.default
@@ -7,7 +7,7 @@
 --enable_optimizer=true
 
 ########## logging ##########
-# The directory to host logging files, which must already exists
+# The directory to host logging files
 --log_dir=logs
 # Log level, 0, 1, 2, 3 for INFO, WARNING, ERROR, FATAL respectively
 --minloglevel=0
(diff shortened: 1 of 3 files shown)


conf/nebula-graphd.conf.default
```diff
@@ -7,7 +7,7 @@
 --enable_optimizer=true
 
 ########## logging ##########
-# The directory to host logging files, which must already exists
```
Shylock-Hg, line 10: But after this PR, it's optional.

yixinglu (approved):

CPWstatic (approved):

conf/nebula-graphd.conf.default
panda-sheep: after this pr, if start the service with a script, the script will create a log directory, if don’t start the service with a script, the service creates a directory.

PR #893, opened 2021-03-29
Fix version bug when packaging

yixinglu:
introduced in #753

### DIFF
--- a/package/package.sh
+++ b/package/package.sh
@@ -19,11 +19,13 @@ package_one=ON
 strip_enable="FALSE"
 usage="Usage: ${0} -v <version> -n <ON/OFF> -s <TRUE/FALSE> -b <BRANCH> -g <ON/OFF>"
 project_dir="$(cd "$(dirname "$0")" && pwd)/.."
-build_dir=${project_dir}/build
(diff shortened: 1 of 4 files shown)


laura-ding (approved):

jievince (approved):

package/package.sh
```diff
@@ -19,11 +19,13 @@ package_one=ON
 strip_enable="FALSE"
 usage="Usage: ${0} -v <version> -n <ON/OFF> -s <TRUE/FALSE> -b <BRANCH> -g <ON/OFF>"
 project_dir="$(cd "$(dirname "$0")" && pwd)/.."
-build_dir=${project_dir}/build
+modules_dir=${project_dir}/modules
+storage_dir=${modules_dir}/storage
```
laura-ding, line 23: It will cause confusion when the `ENABLE_BUILD_STORAGE` is ON.

yixinglu: graph will always need not to build storage, so it`s safe to clone storage here.

laura-ding: if someone uses his code to use the package.sh, the storage will not clone again. and it use the same common.

PR #901, opened 2021-03-30
fix agg executor

czpmango:
Fixes [https://jira.nebula-graph.io/browse/NG-501](https://jira.nebula-graph.io/browse/NG-501)

Depends on [https://github.com/vesoft-inc/nebula-common/pull/480](https://github.com/vesoft-inc/nebula-common/pull/480)

### DIFF
--- a/src/executor/query/AggregateExecutor.cpp
+++ b/src/executor/query/AggregateExecutor.cpp
@@ -27,6 +27,32 @@ folly::Future<Status> AggregateExecutor::execute() {
     QueryExpressionContext ctx(ectx_);
 
     std::unordered_map<List, std::vector<std::unique_ptr<AggData>>, std::hash<nebula::List>> result;
+
+    // generate default result when input dataset is empty
+    if (UNLIKELY(!iter->valid())) {
+        List defaultValues;
+        bool allAggItems = true;
+        for (size_t i = 0; i < groupItems.size(); ++i) {
+            auto* item = groupItems[i];
+            if (UNLIKELY(item->kind() != Expression::Kind::kAggregate)) {
+                allAggItems = false;
+                break;
+            }
+            AggData aggData;
+            static_cast<AggregateExpression*>(item)->apply(&aggData, Value::kNullValue);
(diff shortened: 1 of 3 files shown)


src/executor/query/AggregateExecutor.cpp
```diff
@@ -27,6 +27,32 @@ folly::Future<Status> AggregateExecutor::execute() {
     QueryExpressionContext ctx(ectx_);
 
     std::unordered_map<List, std::vector<std::unique_ptr<AggData>>, std::hash<nebula::List>> result;
+
+    // generate default result when input dataset is empty
+    if (UNLIKELY(!iter->valid())) {
```
Shylock-Hg, line 32: I think it's extra code. If your agg function don't apply at all it will result the default value of course.

yixinglu (commented):
Add test cases

src/executor/query/AggregateExecutor.cpp
czpmango: Be consistent with cypher:
```
neo4j@neo4j> match (v:player) where v.age<(-3) return sum(v.age);
+------------+
| sum(v.age) |
+------------+
| 0          |
+------------+

```

czpmango:
> Add test cases

Already Add.

src/executor/query/AggregateExecutor.cpp
Shylock-Hg: I means your agg function should result default value if don't apply at all.

czpmango: If the aggregate function is not applied, then additional interfaces or data structures are required to store the default values ​​of various functions, which is contradictory to the design of `AggregateManager`. 

If you apply the function, you need to reallocate memory of `AggData`. bcz, `AggData` is dynamically allocated and managed with iteration in the for block.

CPWstatic (dismissed):

yixinglu (approved):

CPWstatic (approved):

nevermore3 (approved):

PR #902, opened 2021-03-30
There is no tag id while execute show tags

MMyheart:
### What changes were proposed in this pull request?
1. SHOW TAGS returns add column ID
2. SHOW EDGES returns add column ID
3. SHOW TAG INDEXES returns add column ID
4. SHOW EDGE INDEXES returns add column ID

### Why are the changes needed?
1. We use offline generation of the underlying storage file (sst) to import data to the NebuleGraph cluster，which depends on the id of the schema。
2. SHOW TAGS only return tag name
image
3. now SHOW TAGS return id and name
image
4. show edges、show tag indexes、show edge indexes as same as show tags

### Will break the compatibility? How if so?
1. Incompatible with old logic

### Does this PR introduce any user-facing change?
1. no

### How was this patch tested?
1. all test passed

### Checklist

- [ ] I've informed the technical writer about the documentation change if necessary
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above
- [x] I've run the tests to see all new and existing tests pass

### DIFF
--- a/src/executor/maintain/EdgeExecutor.cpp
+++ b/src/executor/maintain/EdgeExecutor.cpp
@@ -94,14 +94,15 @@ folly::Future<Status> ShowEdgesExecutor::execute() {
             auto edgeItems = std::move(resp).value();
 
             DataSet dataSet;
-            dataSet.colNames = {"Name"};
-            std::set<std::string> orderEdgeNames;
+            dataSet.colNames = {"ID", "Name"};
+            std::set<std::pair<EdgeType, std::string>> orderEdgeIdNames;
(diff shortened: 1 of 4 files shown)


yixinglu:
Hi @MMyheart Thanks for your contribution!

Some tips to help you to fix the failed test cases, firstly you should read the [tests/README.md](https://github.com/vesoft-inc/nebula-graph/blob/master/tests/README.md) to check how to run all test cases. According to your changes, the cases related to `SHOW TAGS/EDGES` is what you should mind.

PR #905, opened 2021-03-31
Group default conflict check

darionyaphet:
Group default as the global resource should not created


Shylock-Hg:
I think you should change the checking in storage, if group name is `default`, don't check it exists.

darionyaphet:
> I think you should change the checking in storage, if group name is `default`, don't check it exists.

That's not the truth You should not create the `default` group. Not Check

src/executor/admin/GroupExecutor.cpp
```diff
@@ -14,6 +14,11 @@ namespace graph {
 folly::Future<Status> AddGroupExecutor::execute() {
     SCOPED_TIMER(&execTime_);
     auto *agNode = asNode<AddGroup>(node());
+    if (agNode->groupName() == "default") {
```
yixinglu, line 17: Maybe we should check whether the group name is default in parser or validator rather than executor here.

darionyaphet: reasonable

yixinglu (approved):
LGTM

src/validator/AdminValidator.cpp
```diff
@@ -92,8 +92,12 @@ Status CreateSpaceValidator::validateImpl() {
                 } else {
                     spaceDesc_.set_isolation_level(meta::cpp2::IsolationLevel::DEFAULT);
                 }
+                break;
             }
             case SpaceOptItem::GROUP_NAME: {
+                if (item->getGroupName() == "default") {
```
panda-sheep, line 98: Use `folly::toLowerAscii` to deal with, and then compare ？

darionyaphet: Because group name case sensitive, so it's only limit the `default`

yixinglu:
You could add some tests in feature file to test the failure error.

src/validator/AdminValidator.cpp
panda-sheep: `default `as the group name, should ignore the case sensitive， WDYT

yixinglu (dismissed):

yixinglu (dismissed):

yixinglu (approved):

panda-sheep (approved):

PR #907, opened 2021-03-31
Fix the space group don't handled. And fix the atomic_edge type to bool.

Shylock-Hg:
Require https://github.com/vesoft-inc/nebula-storage/pull/414


yixinglu (approved):

darionyaphet:
duplicate with https://github.com/vesoft-inc/nebula-graph/pull/843

laura-ding (approved):

yixinglu (approved):

PR #909, opened 2021-03-31
Unify the fetch/lookup/subgraph yield

laura-ding:
depends on https://github.com/vesoft-inc/nebula-common/pull/479


Shylock-Hg:
According to out previous discussion the fetch/lookup will return the vertex/edge structure?

laura-ding:
> According to out previous discussion the fetch/lookup will return the vertex/edge structure?

The fetch can return vertex/edge or properties. the lookup no requires return vertex/edge, only unify the vertex and edge's key.

src/parser/parser.yy
```diff
@@ -825,6 +832,21 @@ input_prop_expression
     | INPUT_REF DOT STAR {
         $$ = new InputPropertyExpression(new std::string("*"));
     }
+    | INPUT_REF DOT VID_PROP {
```
Shylock-Hg, line 835: I don't think it's useful. For input properties , user must rename then use it.

src/parser/parser.yy
```diff
@@ -843,6 +868,21 @@ var_prop_expression
     | VARIABLE DOT STAR {
         $$ = new VariablePropertyExpression($1, new std::string("*"));
     }
+    | VARIABLE DOT VID_PROP {
```
Shylock-Hg, line 871: ditto.

src/util/OutputColName.h
```diff
@@ -0,0 +1,17 @@
+/* Copyright (c) 2021 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
```
nevermore3, line 6: better put these definitions to base.h

src/validator/LookupValidator.cpp
```diff
@@ -133,26 +137,61 @@ Status LookupValidator::prepareYield() {
...
             if (ret == meta::cpp2::PropertyType::UNKNOWN) {
                 return Status::SemanticError(
-                    "Column %s not found in schema %s", colName.c_str(), from_.c_str());
+                        "Column %s not found in schema %s", colName.c_str(), from_.c_str());
             }
             returnCols_->emplace_back(colName);
             idxScanColNames_.emplace_back(from_ + "." + colName);
             colNames_.emplace_back(deduceColName(newYieldColumns_->back()));
             outputs_.emplace_back(colNames_.back(), SchemaUtil::propTypeToValueType(ret));
+        } else if (col->expr()->kind() == Expression::Kind::kVidExpr) {
```
nevermore3, line 160: better use switch

src/parser/parser.yy
```diff
@@ -1837,7 +1877,6 @@ find_path_sentence
         s->setTo($6);
         s->setOver($7);
         s->setStep($8);
-        /* s->setWhere($9); */
         $$ = s;
     }
     | KW_FIND KW_NOLOOP KW_PATH opt_with_properites from_clause to_clause over_clause find_path_upto_clause
```
nevermore3, line 1882: find path  yield path is missing

src/validator/FindPathValidator.cpp
```diff
@@ -206,7 +206,7 @@ Status FindPathValidator::allPairPaths() {
 
     auto* conjunct = ConjunctPath::make(
         qctx_, forward, backward, ConjunctPath::PathKind::kAllPaths, steps_.steps);
-    conjunct->setColNames({"_path"});
+    conjunct->setColNames({kPathStr});
```
nevermore3, line 209: conjunct‘s column name is not the output column name, it is used internally, if changed,   datacollect executor should be changed Simultaneously

src/parser/parser.yy
```diff
@@ -1910,8 +1948,16 @@ both_in_out_clause
     | KW_BOTH over_edges { $$ = new BothInOutClause($2, BoundClause::BOTH); }
 
 get_subgraph_sentence
-    : KW_GET KW_SUBGRAPH step_clause from_clause in_bound_clause out_bound_clause both_in_out_clause {
-        $$ = new GetSubgraphSentence($3, $4, $5, $6, $7);
+    : KW_GET KW_SUBGRAPH step_clause from_clause in_bound_clause out_bound_clause both_in_out_clause yield_clause {
```
nevermore3, line 1951: when $8!= nullptr ,  how to deal with it

nevermore3:
could you split this PR into several PRs according to different module

src/util/OutputColName.h
laura-ding: It's just used by Graphd

src/parser/parser.yy
laura-ding: The path is not done by this PR

laura-ding:
> could you split this PR into several PRs according to different module

In fact, the code is not much, the most modified is the tests, separated, I will modify the test again.

PR #923, opened 2021-04-07
Fix docker tag name according to action event

yixinglu:
as title.


jievince (dismissed):

laura-ding (dismissed):

yixinglu:
reference vesoft-inc/nebula-storage#425

PR #928, opened 2021-04-07
Rewrite relational expression

Aiee:
1. Move evaluable operands to the right side of the relational expression:
    `v.age + 5 > 40   =>   v.age > 40 -5`

2. if contains any operand that is `null` , rewrite to `null` :
    `expr = null` and `expr != null` and `expr > null` => `null`

3. in the case of `relEq` :
    `expr = true` and `expr != false` to `expr`
    `expr = false` and `expr != true` to `!expr`
    `true = true` and `false = false` to `true`
    `false = true` and `true = false` to `false`

Depends on https://github.com/vesoft-inc/nebula-common/pull/489
TODO:
- [x] Add tck case
- [x] Support to apply the rewrite rule recursively
- [x] Add `filterTransform()` that applies all expression rewrite strategies.


nevermore3 (approved):

src/visitor/test/RewriteRelExprVisitorTest.cpp
```diff
@@ -0,0 +1,105 @@
...
+                           constantExpr(3)),
+                   constantExpr(40)));
+        auto res = pool.add(ExpressionUtils::rewriteRelExpr(expr));
+        auto expected = pool.add(
+            ltExpr(laExpr("v", "age"),
+                   minusExpr(addExpr(minusExpr(constantExpr(40), constantExpr(3)), constantExpr(2)),
+                             constantExpr(1))));
+        ASSERT_EQ(*res, *expected) << res->toString() << " vs. " << expected->toString();
+    }
+}
```
Shylock-Hg, line 102: Add nested relational expression case.

src/util/ExpressionUtils.h
```diff
@@ -155,12 +156,20 @@ class ExpressionUtils {
                                           Expression::Kind::kMod});
     }
 
-    // **Expression Transformation**
+    // Rewrite relational expression, gather evaluable expressions to one side
+    static Expression* rewriteRelExpr(const Expression* expr, ObjectPool* pool);
+    static Expression* rewriteRelExprHelper(const Expression* expr,
+                                            std::unique_ptr<Expression>& relRightOperandExpr);
+
+    static Expression* moveEvaluableExprToRight(const Expression* expr);
```
Aiee, line 164: remove unimplemented function

src/visitor/test/RewriteRelExprVisitorTest.cpp
Aiee: nested relational expressions are rarely used as filters and this PR hasn't addressed this.

Shylock-Hg:
2. I think this is overlap with fold constant expression.
3. What's the benefit?

Aiee:
> 1. What's the benefit?

1. It is not overlapping constant fold. Currently, we cannot fold constants within a relational expression if they are on the same side as non-evaluable expressions. i,g, `v.age + 1 + 2 < 40'
2. It reduces the calculation required in the constant fold when there is a boolean literal in the relational expression. In constant fold, we need to `eval()` both sides of the relational expression then `eval()` the relational expression itself. With the rewrite, once we detect one side of relExpr is a boolean literal, we can rewrite the relExpr directly without further calculation.

src/util/ExpressionUtils.cpp
```diff
@@ -315,6 +428,25 @@ Expression::Kind ExpressionUtils::getNegatedRelExprKind(const Expression::Kind k
...
 
+Expression::Kind ExpressionUtils::getNegatedArithmeticType(const Expression::Kind kind) {
+    switch (kind) {
+        case Expression::Kind::kAdd:
+            return Expression::Kind::kMinus;
+        case Expression::Kind::kMinus:
+            return Expression::Kind::kAdd;
+        case Expression::Kind::kMultiply:
+            return Expression::Kind::kDivision;
+        case Expression::Kind::kDivision:
```
yixinglu, line 439: Test the relational expression which has arithmetic expression children.

tests/tck/features/match/SeekById.intVid.feature
```diff
@@ -244,13 +244,15 @@ Feature: Match seek by id
...
+
+  Scenario: with arithmetic
     When executing query:
       """
       MATCH (v)
       WHERE (id(v) + 1) == hash('James Harden')
       RETURN v.name AS Name
       """
-    Then a SemanticError should be raised at runtime:
+    Then the execution should be successful
```
yixinglu, line 255: should add the expected results

Aiee: Since we don't know the exact value of `hash('James Harden')-1`, there may or may not be a matching result.

yixinglu (approved):

jievince (approved):

PR #929, opened 2021-04-07
Insert if not exist

nevermore3:
depend on vesoft-inc/nebula-common#473
vesoft-inc/nebula-storage#413

### DIFF
--- a/tests/tck/features/insert/InsertIfNotExists.feature
+++ b/tests/tck/features/insert/InsertIfNotExists.feature
@@ -32,8 +32,8 @@ Feature: Insert string vid of vertex and edge
       FETCH PROP ON person "Conan" YIELD person.age as age
       """
     Then the result should be, in any order, with relax comparison:
-      | age |
-      | 10  |
+      | VertexID | age |
+      | "Conan"  | 10  |
     # insert vertex if not exists
     When try to execute query:
       """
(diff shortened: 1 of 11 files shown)


CPWstatic (approved):

avently:
Hello.
Could you include in this PR (or later) an ability to NOT insert vertex/edge IF one or more specified params already exist in a database?

Example:
Tag User(username, email, name, age)

Multiple users can have equal age but username and email should be unique across all Users.

Something like:
INSERT VERTEX user(username, email, age) IF NOT EXISTS (username, email) VALUES ....

nevermore3:
> Hello.
> Could you include in this PR (or later) an ability to NOT insert vertex/edge IF one or more specified params already exist in a database?
> 
> Example:
> Tag User(username, email, name, age)
> 
> Multiple users can have equal age but username and email should be unique across all Users.
> 
> Something like:
> INSERT VERTEX user(username, email, age) IF NOT EXISTS (username, email) VALUES ....

This is a very good suggestion. I think it involves a unique index.  Check whether it is unique before inserting data, We need to plan overall. Thank you for your suggestion.

yixinglu (dismissed):
LGTM

yixinglu (approved):

tests/tck/features/insert/InsertIfNotExists.feature
```diff
@@ -0,0 +1,170 @@
...
+      """
+      FETCH PROP ON person "Conan" YIELD person.age as age
+      """
+    Then the result should be, in any order, with relax comparison:
+      | VertexID | age |
+      | "Conan"  | 10  |
+    # insert vertex if not exists
+    When executing query:
+      """
+      INSERT VERTEX IF NOT EXISTS person(name, age) VALUES "Conan":("Conan", 20)
```
czpmango, line 46: After the execution is successful, is there any messege to indicate that the current data has not been successfully inserted？

czpmango (approved):

CPWstatic (approved):

PR #939, opened 2021-04-12
Correct the latency histogram factor.

Shylock-Hg:


yixinglu (commented):
Why the min value of histogram is set to 2000?

Shylock-Hg:
> Why the min value of histogram is set to 2000?

I don't the latency will less than 2000us.

dutor:
> > Why the min value of histogram is set to 2000?
> 
> I don't think the latency will less than 2000us.

It definitely will.

suggestion: 100000(buckets), 0(ming), 10000000(max)

dutor:
> > > Why the min value of histogram is set to 2000?
> > 
> > 
> > I don't think the latency will less than 2000us.
> 
> It definitely will.
> 
> suggestion: 100000(buckets), 0(ming), 10000000(max)

If concerning the memory consumption, they could be set to 50000, 0, 5000000

Shylock-Hg:
> > > Why the min value of histogram is set to 2000?
> > 
> > 
> > I don't think the latency will less than 2000us.
> 
> It definitely will.
> 
> suggestion: 100000(buckets), 0(ming), 10000000(max)

Set to these.

laura-ding:
> > > > Why the min value of histogram is set to 2000?
> > > 
> > > 
> > > I don't think the latency will less than 2000us.
> > 
> > 
> > It definitely will.
> > suggestion: 100000(buckets), 0(ming), 10000000(max)
> 
> If concerning the memory consumption, they could be set to 50000, 0, 5000000

It is better to enables it to let the user configure?

PR #940, opened 2021-04-12
Feature/multi lang label

Shylock-Hg:
Compatible with opencypher


nevermore3 (dismissed):

jievince (dismissed):

src/parser/scanner.lex
```diff
@@ -36,6 +36,7 @@ IS_EMPTY                    (IS{blanks}EMPTY)
 IS_NOT_EMPTY                (IS{blanks}NOT{blanks}EMPTY)
 
 LABEL                       ([a-zA-Z][_a-zA-Z0-9]*)
+EXTEND_LABEL                ([^`]+)
```
CPWstatic, line 39: Would this lead to that users could write out any forms of identifiers, like `***`, `_____vid`？

Shylock-Hg: It's allowed, maybe we should check that property name shouldn't be the reserved name when create schema.

PR #952, opened 2021-04-14
Add storage_client_timeout_ms and ws_meta_http_port into conf file

laura-ding:


conf/nebula-graphd.conf.production
```diff
@@ -59,6 +59,10 @@
 --ws_http_port=19669
 # HTTP2 service port
 --ws_h2_port=19670
+# storage client timeout
+--storage_client_timeout_ms=60000
+# Port to listen on Meta with HTTP protocol, it corresponds to ws_http_port in metad's configuration file
+--ws_meta_http_port=19559
```
yixinglu, line 65: What's the usage of this port in graphd?

laura-ding: download and ingest

yixinglu (approved):
Why download and ingest use http port to communicate between metad and graphd? We need to discuss this topic offline. Now let it go.

@darionyaphet @CPWstatic @dutor

darionyaphet:
Historical reasons

Maybe move forward to jobmanager in the future

jievince (approved):

Issue #869, opened 2021-03-24
[bug] is itlegal to reference a not exist property (a.k.a property name typo)


whitewum:
```
match (s1:stocks )-[e:relation]->(s2:stocks) where e.weight > 0.1 return s1.name, e.weight, s2.name
```

I should type `e.weight` but typo `e.weigth`

the return column of `e.weigth` is __NULL__. 

When I create the edge, I set relation.weight NOT NULL. 

SO I confuse a sec, and find out that typo.

But, Is it legal to return a not existed alias or property name?

CPWstatic:
As far as i known, it's legal in openCypher/neo4j.

HarrisChu:
https://discuss.nebula-graph.com.cn/t/topic/4413

It's difficult to find the typo error in a complex query, and the result data set may be confused.

```
# e.harris is NULL, so these is no data set.
(root@nebula) [nba]>  match (v:player)- [e] -() where e.harris > 1   return v  limit 1
Empty set (time spent 10305/19358 us)

Tue, 15 Jun 2021 14:21:09 CST
```

CPWstatic:
As I replied earlier that it's legal in openCypher/neo4j. I'll close this issue first, let us know if you have a further concern.

PR #959, opened 2021-04-15
add aliases schema & check it

nevermore3:
refer to https://github.com/vesoft-inc/nebula-graph/issues/869

### DIFF
--- a/tests/tck/features/match/Base.IntVid.feature
+++ b/tests/tck/features/match/Base.IntVid.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 vesoft inc. All rights reserved.
+# Copyright (c) 2021 vesoft inc. All rights reserved.
 #
 # This source code is licensed under Apache 2.0 License,
 # attached with Common Clause Condition 1.0, found in the LICENSES directory.
@@ -369,8 +369,7 @@ Feature: Basic match
       """
       MATCH (:player{name:"Tony Parker"})-[r]->(m) where exists(m.likeness) return r, exists({a:12}.a)
       """
-    Then the result should be, in any order, with relax comparison:
-      | r | exists({a:12}.a) |
+    Then a SemanticError should be raised at runtime: Not found prop `likeness'
     When executing query:
       """
       match (:player{name:"Tony Parker"})-[r]->(m) where exists({abc:123}.abc) return r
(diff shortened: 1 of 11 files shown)


tests/tck/features/match/SeekByTag.intVid.feature
```diff
@@ -18,9 +18,7 @@ Feature: Match seek by tag
       MATCH (v:bachelor)
       RETURN id(v) AS vid, v.age AS age
       """
-    Then the result should be, in any order, and the columns 0 should be hashed:
-      | vid          | age |
-      | 'Tim Duncan' | 42  |
+    Then a SemanticError should be raised at runtime: Not found prop `age'
```
Shylock-Hg, line 21: It's runable.

nevermore3: bachelor has no  age attribute， how could it be workable

Shylock-Hg: The vertex has it. I think you could check the properties in this space and report error if don't exists.

yixinglu (approved):

czpmango:
Will the behavior of schema checking be consistent with neo4j or follow the cypher standard?

czpmango:
IMO, this may conflict with the original intention of `cypher` language design.

nevermore3:
> Will the behavior of schema checking be consistent with neo4j or follow the cypher standard?

We are a strong schema type, but cyhper is not, so I think we can refer to mysql's type checking mechanism

czpmango:
> > Will the behavior of schema checking be consistent with neo4j or follow the cypher standard?
> 
> We are a strong schema type, but cyhper is not, so I think we can refer to mysql's type checking mechanism

`nGql` or `match`?

nevermore3:
> > > Will the behavior of schema checking be consistent with neo4j or follow the cypher standard?
> > 
> > 
> > We are a strong schema type, but cyhper is not, so I think we can refer to mysql's type checking mechanism
> 
> `nGql` or `match`?

both

tests/tck/features/match/Base.IntVid.feature
```diff
@@ -406,6 +405,38 @@ Feature: Basic match
       | [:like "Tony Parker"->"Manu Ginobili" @0 {likeness: 95}]                             |
       | [:like "Tony Parker"->"Tim Duncan" @0 {likeness: 95}]                                |
 
+  Scenario: check Alias attribute
```
czpmango, line 408: Could you add test case like this:
```
MATCH (a:player)-[b]-> (c) where c.age>45 or c.name >= "Heat" or a.noexist>2.3 return c
```

czpmango:
> > > > Will the behavior of schema checking be consistent with neo4j or follow the cypher standard?
> > > 
> > > 
> > > We are a strong schema type, but cyhper is not, so I think we can refer to mysql's type checking mechanism
> > 
> > 
> > `nGql` or `match`?
> 
> both

This is quite different from cypher. Do we still need to ensure that the query result is compatible with cypher? How do you deal with short-circuit logic?

tests/tck/features/match/Base.IntVid.feature
nevermore3: already exist in RelationalExpr.feature

tests/tck/features/match/SeekByTag.feature
```diff
@@ -18,9 +18,7 @@ Feature: Match seek by tag
       MATCH (v:bachelor)
       RETURN id(v) AS vid, v.age AS age
       """
-    Then the result should be, in any order:
-      | vid          | age |
-      | 'Tim Duncan' | 42  |
+    Then a SemanticError should be raised at runtime: Not found prop `age'
```
yixinglu, line 21: `(v:bachelor)` doesn't mean vertex `v` only has one tag named `bachelor`. It means `v` must have the tag `bachelor` and maybe have multiple other tags.

yixinglu: So this limitation check is not a reasonable way.

Issue #846, opened 2021-03-19
 Non-self nesting aggregate Expression unsupported


czpmango:
Different from `count(avg(v.age))`, similar to `count(v.age)+avg(v.age)` should be valid.

```
(czp@nebula) [nba]> match (v:player)  return v.age,count(v.age)+avg(v.age)
[ERROR (-12)]: SemanticError: Aggregate function nesting is not allowed: `(COUNT(v.age)+AVG(v.age))'
```

PR #968, opened 2021-04-19
aggregate enhancement

czpmango:
support operation between aggs.

Fixes [https://jira.nebula-graph.io/browse/NG-516](https://jira.nebula-graph.io/browse/NG-516)
Fixes [https://github.com/vesoft-inc/nebula-graph/issues/846](https://github.com/vesoft-inc/nebula-graph/issues/846)


src/validator/GroupByValidator.cpp
```diff
@@ -34,38 +34,55 @@ Status GroupByValidator::validateYield(const YieldClause* yieldClause) {
...
+            }
         }
-        auto colExpr = col->expr();
+
         // collect exprs for check later
         if (colExpr->kind() == Expression::Kind::kAggregate) {
             auto* aggExpr = static_cast<AggregateExpression*>(colExpr);
             NG_RETURN_IF_ERROR(ExpressionUtils::checkAggExpr(aggExpr));
-        } else {
+        } else if (!ExpressionUtils::isEvaluableExpr(colExpr)) {
```
CPWstatic, line 67: Why we have this restriction？

tests/tck/features/aggregate/Agg.feature
```diff
@@ -384,6 +384,18 @@ Feature: Basic Aggregate and GroupBy
       | dst             | count |
       | "Tony Parker"   | 1     |
       | "Manu Ginobili" | 1     |
+    When executing query:
+      """
+      GO FROM "Tim Duncan" OVER like YIELD like._dst AS dst, $$.player.age AS age
+      | GROUP BY $-.dst
+      YIELD
+        $-.dst AS dst,
+        (INT)(sum($-.age)/count($-.age))+avg(distinct $-.age+1)+1 AS statistics
```
CPWstatic, line 393: Do we have some exception cases such as `sum(count())`?

czpmango: Yep, in Agg.feature `error check` scenario.

src/validator/GroupByValidator.cpp
czpmango: Avoid constant expressions as `groupKey` when doing implicit groupby.

src/validator/MatchValidator.cpp
```diff
@@ -685,30 +685,35 @@ Status MatchValidator::validateGroup(YieldClauseContext &yieldCtx) const {
...
+                                                 colExpr->toString().c_str());
+                }
+
+                yieldCtx.groupItems_.emplace_back(
+                    yieldCtx.qctx->objPool()->add(agg->clone().release()));
+
                 yieldCtx.needGenProject_ = true;
+                yieldCtx.aggOutputColumnNames_.emplace_back(agg->toString());
+            }
+            if (aggs.size()) {
```
Shylock-Hg, line 708: !aggs.empty()

czpmango: ACK.

yixinglu (commented):
Why not consider to merge the aggregation validation for groupby and match validators?

czpmango:
> Why not consider to merge the aggregation validation for groupby and match validators?

There are many differences between ngql and cypher, such as:
 - ngql's schema check is stricter than cypher
 - cypher retains more execution information than ngql when doing Multi-step-matching
 - cypher only support implicit-groupby, ngql's groupby is more like mysql
 - ...
 
Some of the above limitations cannot be circumvented at the implementation layer.

Although the implementation is separated at present, the common behavior has been abstracted as util-functions as possible to ensure the DRY principle, such as some expression-related processing.

However, it is still possible to do some refactoring to unify their implementation in the future, but it can not be merged now.

yixinglu (approved):

tests/tck/features/aggregate/Agg.feature
```diff
@@ -418,6 +430,17 @@ Feature: Basic Aggregate and GroupBy
       | 45  | 13    |
       | 46  | 2     |
       | 47  | 3     |
+    When executing query:
```
Shylock-Hg, line 433: Do you have a case about nested agg call.

czpmango: Same as https://github.com/vesoft-inc/nebula-graph/pull/968#discussion_r616303222 ?

jievince (approved):

PR #986, opened 2021-04-26
Modify errorcode

laura-ding:
depends on 
https://github.com/vesoft-inc/nebula-common/pull/517
https://github.com/vesoft-inc/nebula-storage/pull/448

it just uses the new errorCode. It needs to modify the Metaclient's interfaces to return the Metad's errorCode. Then the graphd will unifies errorCode processing.

### DIFF
--- a/src/service/GraphService.cpp
+++ b/src/service/GraphService.cpp
@@ -4,10 +4,13 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
+#include <thrift/lib/cpp/util/EnumUtils.h>
+
 #include "common/base/Base.h"
 #include "common/time/Duration.h"
 #include "common/encryption/MD5Utils.h"
 #include "common/clients/storage/GraphStorageClient.h"
+#include "common/interface/gen-cpp2/common_constants.h"
 #include "service/GraphService.h"
 #include "service/RequestContext.h"
 #include "service/GraphFlags.h"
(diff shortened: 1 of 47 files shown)


src/service/GraphService.cpp
```diff
@@ -91,60 +97,36 @@ GraphService::future_execute(int64_t sessionId, const std::string& query) {
...
-            return "Fail to connect";
-        case ErrorCode::E_BAD_PERMISSION:
-            return "Bad permission";
-        case ErrorCode::E_USER_NOT_FOUND:
-            return "User not found";
-        case ErrorCode::E_TOO_MANY_CONNECTIONS:
-            return "Too many connections";
-        case ErrorCode::E_PARTIAL_SUCCEEDED:
-            return "Partial results";
+    auto resultIter = findIter->second.find(nebula::cpp2::Language::L_EN);
```
CPWstatic, line 111: The language should be configurable, right?

src/service/GraphService.cpp
```diff
@@ -38,15 +41,18 @@ folly::Future<AuthResponse> GraphService::future_authenticate(
...
+        onHandle(ctx, nebula::cpp2::ErrorCode::SUCCEEDED);
     } else {
-        onHandle(ctx, ErrorCode::E_BAD_USERNAME_PASSWORD);
+        auto code = auth(username, password);
+        if (code == nebula::cpp2::ErrorCode::SUCCEEDED) {
+            auto roles = queryEngine_->metaClient()->getRolesByUserFromCache(username);
+            for (const auto& role : roles) {
+                ctx.session()->setRole(role.get_space_id(), role.get_role_type());
+            }
+            onHandle(ctx, nebula::cpp2::ErrorCode::SUCCEEDED);
```
monadbobo, line 52: Here you can use onHandle(ctx, code) directly.

PR #989, opened 2021-04-27
add go step limit OR sample (validate)

nevermore3:
support setting different limit or sample for each step in GO
For example:
GO 3 STEPS FROM <vid> OVER <e> WHERE [filter1, filter2, filter3] YIELD <expr> LIMIT [n1, n2, n3]
GO 3 STEPS FROM <vid> OVER <e> WHERE [filter1, filter2, filter3] YIELD <expr> SAMPLE [n1, n2, n3]

As well as WHERE Clause

### DIFF
--- a/src/validator/GoValidator.cpp
+++ b/src/validator/GoValidator.cpp
@@ -83,18 +83,18 @@ Status GoValidator::validateTruncate(TruncateClause* truncate) {
         return Status::OK();
     }
     random_ = truncate->isSample();
-    auto* expr = truncate->truncate();
-    if (expr->kind() != Expression::Kind::kList) {
-        return Status::SemanticError("`%s' type must be LIST", expr->toString().c_str());
+    auto* tExpr = truncate->truncate();
+    if (tExpr->kind() != Expression::Kind::kList) {
+        return Status::SemanticError("`%s' type must be LIST", tExpr->toString().c_str());
     }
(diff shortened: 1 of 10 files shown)


src/validator/GoValidator.cpp
```diff
@@ -125,6 +126,44 @@ Status GoValidator::validateYield(YieldClause* yield) {
...
+        }
+        return false;
+    };
+    FindVisitor visitor(existNonInteger);
+    tExpr->accept(&visitor);
+    auto res = visitor.results();
+    if (res.size() == 1) {
+        return Status::SemanticError("`%s' must be INT", res.front()->toString().c_str());
+    }
+    return Status::OK();
```
CPWstatic, line 164: We could return error for now since we don't implement it.

nevermore3: ok

Shylock-Hg:
Could you explain the meaning of new syntax?

nevermore3:
> Could you explain the meaning of new syntax?

In the first step of expansion, I only take 10 edges, and in the second step of expansion, I only take 30 edges， I can use the following syntax

go 2 steps from "Tim Duncan" over like limit [10, 30], same with SAMPLE

CPWstatic (approved):

yixinglu (approved):

PR #990, opened 2021-04-28
make findvisitor suitable for more scenarios

nevermore3:
for example
go N steps from <vid> over <e> yield e._dst limit[1, 2, 3]
Check if there is a non-integer in the limit list OR check if the value of list is Less than N
There is no targets


czpmango:
IMO, `FindVisitor` is abstracted to search from a set of targets, and returns true if a target is found. It is not appropriate to modify its original abstract intention just because of its different usage.

In your above scenario, you can do as following:
```
auto finder = [N](const Expression* expr, const std::unordered_set<Value::Type>& targets) -> bool {
    if (expr->kind() == Expression::Kind::kConstant) {
        auto val = static_cast<ConstantExpression>(expr)->value();
        if (targets.find(val.type()) != targets.end() || val > N) {
            return true;
        }
    }
    return false;
};
FindVisitor<Value::Type> visitor(finder, {Value::Type::INT});
const_cast<Expression*>(self)->accept(&visitor);
auto res = visitor.results();
```

nevermore3:
> IMO, `FindVisitor` is abstracted to search from a set of targets, and returns true if a target is found. It is not appropriate to modify its original abstract intention just because of its different usage.
> 
> In your above scenario, you can do as following:
> 
> ```
> auto finder = [N](const Expression* expr, const std::unordered_set<Value::Type>& targets) -> bool {
>     if (expr->kind() == Expression::Kind::kConstant) {
>         auto val = static_cast<ConstantExpression>(expr)->value();
>         if (targets.find(val.type()) != targets.end() || val > N) {
>             return true;
>         }
>     }
>     return false;
> };
> FindVisitor<Value::Type> visitor(finder, {Value::Type::INT});
> const_cast<Expression*>(self)->accept(&visitor);
> if (visitor.found()) {
>     return Status::Error("Invalid limit params.");
> }
> ```

this is a solution,  findVisitor can be used as a basic traversal,  variables can be solved by using lambda capture，no need to use member variables

czpmango (approved):
Good job. 
This is a better abstraction for `findVisitor`.

czpmango:
Good job. 
This is a better abstraction for `findVisitor`.

> > IMO, `FindVisitor` is abstracted to search from a set of targets, and returns true if a target is found. It is not appropriate to modify its original abstract intention just because of its different usage.
> > In your above scenario, you can do as following:
> > ```
> > auto finder = [N](const Expression* expr, const std::unordered_set<Value::Type>& targets) -> bool {
> >     if (expr->kind() == Expression::Kind::kConstant) {
> >         auto val = static_cast<ConstantExpression>(expr)->value();
> >         if (targets.find(val.type()) != targets.end() || val > N) {
> >             return true;
> >         }
> >     }
> >     return false;
> > };
> > FindVisitor<Value::Type> visitor(finder, {Value::Type::INT});
> > const_cast<Expression*>(self)->accept(&visitor);
> > if (visitor.found()) {
> >     return Status::Error("Invalid limit params.");
> > }
> > ```
> 
> this is a solution, But findVisitor can be used as a basic traversal, variables can be solved by using lambda capture，no need to use member variables

ACK.

yixinglu (approved):

PR #998, opened 2021-05-06
Optimize/merge int string cases

Shylock-Hg:
Propose a method to abstract the int/string vid test cases. Now we template the test case for int/string vid.There is only difference between vid type, so we could abstract the difference by the feature example method.This example arguments string/hashed vid.


tests/tck/features/fetch/FetchVertices.strVid.feature
```diff
@@ -1,440 +1,444 @@
...
+  Examples:
+    | space_name  | hash_begin  | hash_end |
+    | nba         |             |          |
+    | nba_int_vid | hash(       |    )     |
 
   Background:
-    Given a graph with space named "nba"
+    Given a graph with space named "<space_name>"
 
   Scenario: Fetch prop on one tag, one vertex
```
yixinglu, line 10: Scenario Outline

Shylock-Hg: Fixed.

nevermore3 (approved):

PR #1015, opened 2021-05-10
fix #1012: Collect values from InputExpression crash the nebula graphd

### LINKED ISSUE #1012: Collect values from InputExpression crash the nebula graphd
following query will crash the graphd

```
UNWIND [1,2,3] AS d RETURN d | YIELD 1 IN COLLECT($-.d)
```

But next case works well:

```
UNWIND [1,2,3] AS d RETURN d | YIELD ANY(l IN COLLECT($-.d) WHERE l==1)
```

response:

```
(user@nebula) [nba]> UNWIND [1,2,3] AS d RETURN d | YIELD ANY(l IN COLLECT($-.d) WHERE l==1)
+--------------------------------------+
| any(l IN COLLECT($-.d) WHERE (l==1)) |
+--------------------------------------+
| true                                 |
+--------------------------------------+
Got 1 rows (time spent 662/1267 us)

Sun, 09 May 2021 13:07:55 CST
```

jievince:
fix #1012 

We don't need to pass a `needVisitedTypes` to agg rewriter because the expression kinds that don't need to be visited have an empty function of body implementation.
```
 void visit(ConstantExpression *) override {}
 void visit(LabelExpression*) override {}
 ...
```
Expression kinds such as kConstant, kLabelProperty are basic kinds, they are impossible to nest an agg

### DIFF
--- a/src/util/ExpressionUtils.cpp
+++ b/src/util/ExpressionUtils.cpp
@@ -139,17 +139,7 @@ Expression *ExpressionUtils::rewriteAgg2VarProp(const Expression *expr) {
 
         return RewriteVisitor::transform(expr,
                                          std::move(matcher),
-                                         std::move(rewriter),
-                                         {Expression::Kind::kFunctionCall,
-                                          Expression::Kind::kTypeCasting,
-                                          Expression::Kind::kAdd,
-                                          Expression::Kind::kMinus,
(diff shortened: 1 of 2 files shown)


yixinglu (approved):
LGTM

czpmango (approved):

PR #1032, opened 2021-05-12
disable variable binding of alternative edge types

czpmango:


src/parser/parser.yy
```diff
@@ -1532,10 +1532,6 @@ match_edge_type_list
         $$ = $1;
         $$->add($3);
     }
-    | match_edge_type_list PIPE COLON name_label {
```
Shylock-Hg, line 1535: It's legal syntax.

czpmango:
```
neo4j@neo4j> match (v:teacher)-[e:like|:serve]-(n) return n;
The semantics of using colon in the separation of alternative relationship types in conjunction with
the use of variable binding, inlined property predicates, or variable length is no longer supported.
Please separate the relationships types using `:A|B|C` instead (line 2, column 18 (offset: 18))
"match (v:teacher)-[e:like|:serve]-(n) return n;"
```

Shylock-Hg: As I known, we don't follow the newest version of open cypher. @CPWstatic Which version we followed?

yixinglu:
This change is not necessary. I close it now.

czpmango:
> This change is not necessary. I close it now.

The colon is used to bind variable, the above syntax will introduce semantic ambiguity when processing relationships (multi-step relationships are currently stored in a list), so cypher is gradually banning it. neo4j will also report errors for these queries.

PR #1035, opened 2021-05-13
move condition to util

nevermore3:


src/util/ExpressionUtils.cpp
```diff
@@ -507,5 +507,32 @@ Expression::Kind ExpressionUtils::getNegatedLogicalExprKind(const Expression::Ki
     }
 }
 
+// ++loopStep <= steps
+Expression* ExpressionUtils::stepCondition(const std::string &loopStep, uint32_t steps) {
```
yixinglu, line 511: it's better to return unique_ptr

nevermore3: done

yixinglu (approved):

CPWstatic (approved):

PR #1072, opened 2021-05-25
Add the cmake option to checkout modules.

Shylock-Hg:


cmake/FetchModule.cmake
```diff
@@ -30,34 +30,36 @@ macro(nebula_fetch_module)
         endif()
     else()
         message(STATUS "Updating from ${module_URL}")
```
laura-ding, line 32: here can be delete

cmake/FetchModule.cmake
```diff
@@ -9,7 +9,7 @@ macro(nebula_fetch_module)
     cmake_parse_arguments(
         module                      # <prefix>
```
laura-ding, line 10: alignment

cmake/FetchModule.cmake
Shylock-Hg: Fixed.

cmake/FetchModule.cmake
Shylock-Hg: Fixed.

laura-ding (approved):

nevermore3 (approved):

PR #1075, opened 2021-05-25
Add TCK cases about NULL comparison and NULL with math functions

randomJoe211:


jievince:
Well done! Could you format the tck tests by `make fmt` in the directory `nebula-graph/tests` at first?

whitewum:
These cases are shown in the doc. So if any feature breaks these cases, you will notice.

yixinglu:
before formating your feature file, you can follow [the tutorial](https://github.com/vesoft-inc/nebula-graph/blob/master/tests/README.md#init-environment) to initialize your environment.

yixinglu (approved):

tests/tck/features/expression/Null.feature
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2020 vesoft inc. All rights reserved.
```
laura-ding, line 1: 2021

laura-ding (approved):

PR #1081, opened 2021-05-26
A bright new scheduler implementation based on BFS and Promise/Future mode.

CPWstatic:
Implementation description:
- use the `breadth first search` to avoid stack overflow which introduces by `DFS`
- use `Promise/Future` to achieve an implementation of asynchronous message notification
- the `Start`, `PassThrough` will not be needed anymore
- as for the performance, i had test for several cases based on [nebula-bench](https://github.com/vesoft-inc/nebula-bench). And the results indicates that this implementation has almost the same performance with the master one.


yixinglu (commented):
Could you give an abstract `Scheduler` interface and let the different concrete scheduler to implement this scheduler interface? because i have also enhance the scheduler implementation, maybe we can provide more schedulers implementation and compare them performance later. WDYT?

CPWstatic:
> Could you give an abstract `Scheduler` interface and let the different concrete schedulers to implement this scheduler interface? because i have also enhance the scheduler implementation locally, maybe we can provide more schedulers implementation and compare them performance later. WDYT?

Done.

yixinglu (approved):

nevermore3 (approved):

PR #1085, opened 2021-05-26
Return all rules when execution plan is not stable

yixinglu:
as title

### DIFF
--- a/src/optimizer/OptGroup.cpp
+++ b/src/optimizer/OptGroup.cpp
@@ -28,6 +28,16 @@ OptGroup *OptGroup::create(OptContext *ctx) {
     return ctx->objPool()->add(new OptGroup(ctx));
 }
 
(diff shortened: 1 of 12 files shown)


src/optimizer/OptGroup.cpp
```diff
@@ -28,6 +28,16 @@ OptGroup *OptGroup::create(OptContext *ctx) {
     return ctx->objPool()->add(new OptGroup(ctx));
 }
 
+void OptGroup::setUnexplored(const OptRule *rule) {
+    auto iter = std::find(exploredRules_.begin(), exploredRules_.end(), rule);
```
monadbobo, line 32: Here you can directly use std::erase+std::remove_if

czpmango (dismissed):

Aiee (approved):

czpmango (approved):
LGTM.

PR #1086, opened 2021-05-26
Fix extract filter visitor bug

yixinglu:


Aiee (dismissed):

monadbobo (approved):
LGTM.

czpmango:
This visitor only serve for `nGql`? Bcz I see `LabelAttributeExpression` is regarded as non-pushable.
If a label in the `cypher` statement is actually the starting point, why can't it be embeded into `IndexScan`?

IMHO, Why not abstract "whether it can be pushed down" as a lambda, and let the caller define it?

CPWstatic (approved):

czpmango (approved):

PR #1094, opened 2021-05-31
Shrink sizeof Value struct

dutor:
Depends on vesoft-inc/nebula-common#549 and vesoft-inc/nebula-storage#473


yixinglu (approved):

bright-starry-sky (approved):
Well done.

PR #1096, opened 2021-05-31
Keep the name of vertex/edge expression.

Shylock-Hg:
Require https://github.com/vesoft-inc/nebula-common/pull/553


src/planner/match/Expand.cpp
```diff
@@ -77,8 +77,8 @@ static Expression* mergePathColumnsExpr(const std::string& lcol, const std::stri
 
 static Expression* buildPathExpr() {
     auto expr = std::make_unique<PathBuildExpression>();
-    expr->add(std::make_unique<VertexExpression>());
-    expr->add(std::make_unique<EdgeExpression>());
+    expr->add(std::make_unique<VertexExpression>("VERTEX"));
```
jievince, line 80: Why need to store the name?

Shylock-Hg: 1. print
2. collect the vertex/edge properties.

PR #1098, opened 2021-06-01
Initialize column names for some plan node according to their dependencies

yixinglu:
To Simplify Project/Dedup/Filter contruction:
1. For dedup/filter, set the input's columns as their output columns by default
2. For project, set the yield columns as its output columns when calling constructor.


nevermore3 (approved):

src/planner/plan/PlanNode.cpp
```diff
@@ -349,6 +349,15 @@ std::ostream& operator<<(std::ostream& os, PlanNode::Kind kind) {
     return os;
 }
 
+SingleInputNode::SingleInputNode(QueryContext* qctx, Kind kind, const PlanNode* dep)
+    : SingleDependencyNode(qctx, kind, dep) {
+    if (dep != nullptr) {
+        readVariable(dep->outputVarPtr());
+    } else {
+        inputVars_.emplace_back(nullptr);
```
laura-ding, line 357: The `releaseSymbols()` directly to use the element of inputVars_, need to add check in `releaseSymbols()`.

czpmango (approved):

src/planner/plan/PlanNode.cpp
yixinglu: Has checked in `releaseSymbols()`

PR #1105, opened 2021-06-03
Catch the exception in each future callback.

CPWstatic:


src/scheduler/AsyncMsgNotifyBasedScheduler.cpp
```diff
@@ -177,15 +202,23 @@ void AsyncMsgNotifyBasedScheduler::runLoop(std::vector<folly::Future<Status>>&&
...
+            }
+            auto status = std::move(t).value();
             auto s = checkStatus(std::move(status));
             if (!s.ok()) {
                 return notifyError(pros, s);
             }
 
-            std::move(execute(loop)).thenValue(
-                [loop, runner, pros = std::move(pros), this](Status loopStatus) mutable {
+            std::move(execute(loop)).thenTry(
```
Shylock-Hg, line 216: I think you could try the `thenError` method of future.

CPWstatic: The `promise` will no longer valid after `thenTry`, the `thenError` is not an option.

laura-ding (approved):

yixinglu (approved):

PR #1113, opened 2021-06-04
Let the test cases reentrant.

Shylock-Hg:


tests/admin/test_configs.py
```diff
@@ -93,19 +93,28 @@ def test_configs(self):
         # update rocksdb
         resp = self.client.execute('''
                                    UPDATE CONFIGS storage:rocksdb_column_family_options={
-                                   max_bytes_for_level_base=1024,
-                                   write_buffer_size=1024,
-                                   max_write_buffer_number=4}
+                                   max_bytes_for_level_base="1024",
+                                   write_buffer_size="1024",
+                                   max_write_buffer_number="4"}
```
yixinglu, line 98: why to use double quotes for config values? it's different from flags in cmdline.

Shylock-Hg: It's just string in the storage configuration file.  https://github.com/vesoft-inc/nebula-storage/blob/c180a9a241767c312a50e6e64a341ef0e3f8a11e/conf/nebula-storaged.conf.production#L81

yixinglu (approved):

Aiee (approved):

PR #1116, opened 2021-06-07
allow user to limit graphd memory usage

czpmango:
Depend on [https://github.com/vesoft-inc/nebula-common/pull/558](https://github.com/vesoft-inc/nebula-common/pull/558)

Performance test verification : [https://github.com/czpmango/jemalloc-test](https://github.com/czpmango/jemalloc-test)

### DIFF
--- a/src/executor/Executor.cpp
+++ b/src/executor/Executor.cpp
@@ -95,6 +95,7 @@
 #include "planner/plan/Query.h"
 #include "service/GraphFlags.h"
 #include "util/ScopedTimer.h"
+#include "common/time/Duration.h"
 
 using folly::stringPrintf;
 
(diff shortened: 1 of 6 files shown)


src/executor/Executor.cpp
```diff
@@ -541,16 +542,50 @@ Status Executor::open() {
...
+        }
+    }
+
+    if (node_->isQueryNode()) {
+        auto mem = MemInfo::make();
+        if (isBigQuery || !skipStatsCollection) {
+            NG_RETURN_IF_ERROR(mem->initProcessMemInfo());
+            ectx_->setLastStatsCollectTime(current_time_ms);
+            if (FLAGS_memory_limit_graphd > 0 &&
+                mem->processResidentInByte() > FLAGS_memory_limit_graphd) {
```
critical27, line 571: We forbid this check by default?

czpmango: Yep, forbidden by default.

`stats.resident` is closer to the physical memory actually allocated by the current process. 
We set up some coarse-grained code stubs to get memory statistics such as in `Executor::open()` and look forward to getting the upper limit of the process physical memory usage to prevent oom as much as possible.

PR #1139, opened 2021-06-16
[poc]add match_clause_with_props_flag

jievince:


src/planner/match/Expand.cpp
```diff
@@ -22,6 +23,9 @@ namespace nebula {
 namespace graph {
 
 static std::unique_ptr<std::vector<VertexProp>> genVertexProps() {
+    if (FLAGS_match_clause_with_props) {
```
nevermore3, line 26: std::make_unique<std::vector<VertexProp>>(1);

src/executor/query/GetVerticesExecutor.cpp
```diff
@@ -36,6 +37,13 @@ folly::Future<Status> GetVerticesExecutor::getVertices() {
                           .finish());
     }
 
+    if (!FLAGS_match_clause_with_props) {
```
Shylock-Hg, line 40: The executor isn't used by match only.

PR #1146, opened 2021-06-17
fix create space on group

tongheihei:
create space group_test2(partition_num=9,replica_factor=1) ON group1 .
This PR can make sure that the  'group_test2' space is on the storage nodes of group1 rather than all the storage nodes.


src/validator/AdminValidator.cpp
```diff
@@ -108,6 +111,8 @@ Status CreateSpaceValidator::validateImpl() {
         return Status::SemanticError("Group default conflict");
     }
 
+=======
+>>>>>>> fix create space on group
```
Aiee, line 115: Please delete extra lines.

PR #1155, opened 2021-06-19
Request storage in batch

yixinglu:
just for poc test


nevermore3:
use storage multi-thread
https://github.com/vesoft-inc/nebula-storage/pull/503

Issue #1169, opened 2021-06-23
Mistake check of yield expression in yield sentence.


Shylock-Hg:
```
a = go from 'Tim Duncan' over *; yield $a;
a = go from 'Tim Duncan' over *; yield $a[1][1];
```
In yield validator, mistake check the evaluable of expression of yield column.

Shylock-Hg:
@CPWstatic @yixinglu @laura-ding @Aiee @jievince #764 will analysis the lifetime of variable by the inputVars of plan node, so we must set the all variable used by one plan node to its inputVars, Or 
1. Maybe get empty result for the variable generated by plan node output;
2. Variable keep alive until the query finish for the variable manually generated when validating/planning.

Shylock-Hg:
This only crash in debug mode, get null in release.

PR #1174, opened 2021-06-24
Fix the mistake that yield validator mistake check the input/var.

Shylock-Hg:
Close https://github.com/vesoft-inc/nebula-graph/issues/1169


tests/tck/features/expression/Reduce.feature
```diff
@@ -10,16 +10,12 @@ Feature: Reduce
       """
       YIELD reduce(totalNum = 10, n IN range(1, 3) | totalNum + n) AS r
       """
-    Then the result should be, in any order:
-      | r  |
-      | 16 |
+    Then a SemanticError should be raised at runtime: Multiple user defined vars not supported yet.
```
yixinglu, line 13: this check is not acceptable!

Shylock-Hg: Fixed.

tests/tck/features/bugfix/TestYieldConstantAfterPipe.feature
```diff
@@ -15,25 +15,13 @@ Feature: Test yield constant after pipe
...
-      | 1 |
-      | 1 |
-      | 1 |
     When executing query:
       """
       GO FROM "Tim Duncan" OVER * | YIELD 1 AS a WHERE true;
       """
     Then the result should be, in any order:
       | a |
       | 1 |
```
yixinglu, line 24: I want to know what's the right result should be ? @CPWstatic 

Should the number of results be same as the edges from `Tim Duncan` ?

Shylock-Hg: According to the origin code, I guess it may want to set input to the inner var when the all expressions of yield columns are constant. So I change it to this.

tests/tck/features/bugfix/YieldVar.feature
```diff
@@ -0,0 +1,23 @@
...
+      """
+    Then the result should be, in any order:
+      | var00           |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
+      | "Manu Ginobili" |
```
jievince, line 23: The result of $var[row][col] is not a cell?

tests/tck/features/bugfix/TestYieldConstantAfterPipe.feature
jievince: I think it's not compatible with the definition of pipe

tests/tck/features/bugfix/YieldVar.feature
Shylock-Hg: Yes, it's not constant expr.

Shylock-Hg:
Move to https://github.com/vesoft-inc/nebula/pull/2789

PR #1195, opened 2021-07-02
Transfer download and ingest into job system

darionyaphet:


yixinglu (dismissed):

src/parser/parser.yy
```diff
@@ -2769,15 +2768,6 @@ delete_vertex_sentence
     }
     ;
 
-download_sentence
```
Shylock-Hg, line 2772: Maybe we could remain the syntax.

darionyaphet: I'm not sure. Remove `DOWNLOAD` and `INGEST` sentences that maybe could unify the execution semantics.

Shylock-Hg: You could still add the job syntax but keep these for compatible.

PR #1204, opened 2021-07-05
Do not upload killed queries when queries have not sync to meta.

CPWstatic:


nevermore3 (approved):

src/executor/admin/KillQueryExecutor.cpp
```diff
@@ -141,5 +146,29 @@ Status KillQueryExecutor::verifyTheQueriesByMetaInfo(
     }
     return Status::OK();
 }
+
+void KillQueryExecutor::findKillQueriesViaMeta(
```
Aiee, line 150: Change `findKillQueriesViaMeta` -> `findKilledQueriesViaMeta` may be more appropriate.

Aiee (approved):

PR #1213, opened 2021-07-06
Fix use same filters to lookup index

yixinglu:
as title.


Aiee (dismissed):

Aiee (approved):

CPWstatic (approved):
That's great. We might need some UT for some of the utils.

PR #1219, opened 2021-07-08
fix compilation error in some third-party versions

czpmango:


laura-ding (approved):
under glibc2.28, the value of std::set

yixinglu (approved):

PR #1240, opened 2021-07-12
fix filter degradation

czpmango:
To ensure that `stableFilter`(such as `SequentialIter::erase()`) is used only if the `WITH` clause contains `ORDER BY`.

### DIFF
--- a/src/planner/match/WithClausePlanner.cpp
+++ b/src/planner/match/WithClausePlanner.cpp
@@ -53,7 +53,9 @@ Status WithClausePlanner::buildWith(WithClauseContext* wctx, SubPlan& subPlan) {
     }
 
     if (wctx->where != nullptr) {
-        auto wherePlan = std::make_unique<WhereClausePlanner>()->transform(wctx->where.get());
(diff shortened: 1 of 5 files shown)


src/planner/match/WithClausePlanner.cpp
```diff
@@ -53,7 +53,9 @@ Status WithClausePlanner::buildWith(WithClauseContext* wctx, SubPlan& subPlan) {
     }
 
     if (wctx->where != nullptr) {
-        auto wherePlan = std::make_unique<WhereClausePlanner>()->transform(wctx->where.get());
```
jievince, line 56: bool needStableFilter = wctx->order != nullptr

jievince (approved):

src/planner/match/WithClausePlanner.cpp
czpmango: ok.

jievince (approved):

Aiee (approved):

CPWstatic (approved):

PR #1243, opened 2021-07-12
Replace the lastUser by user count.…

Shylock-Hg:
The last user maybe could run simultaneously(multiple dependencies use the same variable or select branch use the same variable), so can't determine which is real last user in timeline. Although now we can't write such query in fact, but it's a potential risk, so replace it by user count.

### DIFF
--- a/src/executor/Executor.cpp
+++ b/src/executor/Executor.cpp
@@ -582,9 +582,9 @@ folly::Future<Status> Executor::error(Status status) const {
 void Executor::drop() {
     for (const auto &inputVar : node()->inputVars()) {
         if (inputVar != nullptr) {
-            // Make sure use the variable happened-before de-increment count
+            // Make sure use the variable happened-before decrement count
             if (inputVar->userCount.fetch_sub(1, std::memory_order_release) == 1) {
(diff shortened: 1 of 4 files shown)


src/executor/Executor.cpp
```diff
@@ -581,16 +582,20 @@ folly::Future<Status> Executor::error(Status status) const {
 void Executor::drop() {
     for (const auto &inputVar : node()->inputVars()) {
         if (inputVar != nullptr) {
-            if (inputVar->lastUser.value() == node()->id()) {
-                    ectx_->dropResult(inputVar->name);
-                    VLOG(1) << "Drop variable " << node()->outputVar();
+            // Make sure use the variable happened-before de-increment count
```
CPWstatic, line 585: What does `de-increment` means?

CPWstatic (commented):
LGTM.

src/executor/Executor.cpp
Shylock-Hg: Typo. Had fixed.

CPWstatic (approved):

czpmango (approved):

PR #1244, opened 2021-07-12
fix bug for insert vertex | edge

bright-starry-sky:
fixed bugs as below:
Repeated vid or edge , for example delete vertex "a", "a"

### DIFF
--- a/src/executor/mutate/DeleteExecutor.cpp
+++ b/src/executor/mutate/DeleteExecutor.cpp
@@ -58,6 +58,9 @@ folly::Future<Status> DeleteVerticesExecutor::deleteVertices() {
     if (vertices.empty()) {
         return Status::OK();
     }
+    std::sort(vertices.begin(), vertices.end());
+    vertices.erase(unique(vertices.begin(), vertices.end()), vertices.end());
+
     auto spaceId = spaceInfo.id;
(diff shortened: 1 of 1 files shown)


src/executor/mutate/DeleteExecutor.cpp
```diff
@@ -58,6 +58,9 @@ folly::Future<Status> DeleteVerticesExecutor::deleteVertices() {
     if (vertices.empty()) {
         return Status::OK();
     }
+    std::sort(vertices.begin(), vertices.end());
+    vertices.erase(unique(vertices.begin(), vertices.end()), vertices.end());
```
yixinglu, line 62: Need not to sort the vids, can use set to dedup them.

bright-starry-sky: > Need not to sort the vids, can use set to dedup them.

if using std::set , still need to convert it from std::set to std::vector. and it's not a good idea to change the parameter type of storage clients.

Shylock-Hg: I think you could add a `dedup` executor directly.

yixinglu: Agree to add dedup plan node in execution plan.

PR #1257, opened 2021-07-15
fix agg filter check in with clause

czpmango:


jievince (approved):

src/parser/parser.yy
```diff
@@ -1390,9 +1390,25 @@ unwind_clause
 
 with_clause
     : KW_WITH yield_columns match_order_by match_skip match_limit where_clause {
+        if ($6 && graph::ExpressionUtils::findAny($6->filter(),{Expression::Kind::kAggregate})) {
+            delete($2);
+            delete($3);
+            delete($4);
+            delete($5);
+            delete($6);
+            throw nebula::GraphParser::syntax_error(@6, "Invalid use of aggregating function in this context.");
```
Shylock-Hg, line 1399: `"Invalid use of aggregating function in where filter`

Shylock-Hg (approved):
LGTM

Aiee (approved):

yixinglu (approved):

src/parser/parser.yy
czpmango: Alias ​​management does not support this writing method, the current error message is consistent with neo4j.

```
neo4j@neo4j> match (v:teacher) with v,count(v.age) as ages where count(v.age) >= 0 return v.age;
+-------+
| v.age |
+-------+
| 1     |
| TRUE  |
| "17"  |
| NULL  |
| NULL  |
| NULL  |
| NULL  |
+-------+

neo4j@neo4j> match (v:teacher) with v where count(v.age) >= 0 return v.age;
Invalid use of aggregating function count(...) in this context (line 2, column 32 (offset: 32))
"match (v:teacher) with v where count(v.age) >= 0 return v.age;"
                                ^
```

PR #1272, opened 2021-07-20
Free unused multiple version result's memory

yixinglu:
for most plan nodes, it's not necessary to store multiple version's results. So this PR lets these plan nodes update the outputs in place to save the memory usage.

### DIFF
--- a/src/executor/Executor.cpp
+++ b/src/executor/Executor.cpp
@@ -592,9 +592,11 @@ void Executor::drop() {
 Status Executor::finish(Result &&result) {
     if (!FLAGS_enable_lifetime_optimize || node()->outputVarPtr()->lastUser.hasValue()) {
         numRows_ = result.size();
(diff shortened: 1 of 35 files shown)


src/executor/Executor.cpp
```diff
@@ -592,9 +592,11 @@ void Executor::drop() {
 Status Executor::finish(Result &&result) {
     if (!FLAGS_enable_lifetime_optimize || node()->outputVarPtr()->lastUser.hasValue()) {
         numRows_ = result.size();
-        ectx_->setResult(node()->outputVar(), std::move(result));
-    } else {
-        VLOG(1) << "Drop variable " << node()->outputVar();
```
Shylock-Hg, line 597: Don't delete the vlog.

yixinglu: I think it's the debug code, need not to put it here when not debugging.

PR #1282, opened 2021-07-22
[OptRule] Collapse continuous Project plan node

czpmango:
Merge consecutive project plan nodes to avoid memory replication, and this may benefit column pruning in some cases.

example: match (v:player) with v, v.age/10 as age, v.name as name, abs(v.age) as abs return age-1 as res

old plan: 
`...->Project(v,age,name,abs)->Project(res)`

plan applying this rule: 
`...->Project(res)` 
 (res=v.age/10-1)

### DIFF
--- a/src/optimizer/rule/CollapseProjectRule.cpp
+++ b/src/optimizer/rule/CollapseProjectRule.cpp
@@ -1,0 +1,121 @@
+/* Copyright (c) 2021 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "optimizer/rule/CollapseProjectRule.h"
+
+#include "optimizer/OptContext.h"
+#include "optimizer/OptGroup.h"
+#include "planner/plan/PlanNode.h"
+#include "planner/plan/Query.h"
+
+using nebula::graph::PlanNode;
+using nebula::graph::QueryContext;
+
+namespace nebula {
+namespace opt {
+
+std::unique_ptr<OptRule> CollapseProjectRule::kInstance =
+    std::unique_ptr<CollapseProjectRule>(new CollapseProjectRule());
+
+CollapseProjectRule::CollapseProjectRule() {
+    RuleSet::QueryRules().addRule(this);
+}
+
+const Pattern& CollapseProjectRule::pattern() const {
+    static Pattern pattern = Pattern::create(graph::PlanNode::Kind::kProject,
+                                             {Pattern::create(graph::PlanNode::Kind::kProject)});
+    return pattern;
+}
+
+StatusOr<OptRule::TransformResult> CollapseProjectRule::transform(
+    OptContext* octx,
+    const MatchedResult& matched) const {
+    const auto* groupNodeAbove = matched.node;
+    const auto* projAbove = groupNodeAbove->node();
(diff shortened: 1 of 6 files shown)


tests/tck/features/optimizer/CollapseProjectRule.feature
```diff
@@ -0,0 +1,54 @@
...
+      | 4    |
+      | 4    |
+      | 4    |
+      | 3    |
+    And the execution plan should be:
+      | id | name        | dependencies | operator info |
+      | 16 | Project     | 14           |               |
+      | 14 | Filter      | 7            |               |
+      | 7  | Project     | 6            |               |
+      | 6  | Project     | 5            |               |
```
yixinglu, line 36: could these projects not be collapsed ?

src/optimizer/rule/CollapseProjectRule.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    // 3. rewrite YieldColumns
+    auto matcher = [&rewriteMap](const Expression* e) -> bool {
+        if (e->kind() != Expression::Kind::kVarProperty &&
+            e->kind() != Expression::Kind::kInputProperty) {
+            return false;
+        }
+        auto& propName = static_cast<const PropertyExpression*>(e)->prop();
+        return rewriteMap[propName];
```
jievince, line 87: What does it mean? I don't get it

tests/tck/features/optimizer/CollapseProjectRule.feature
czpmango: In order to avoid multiple calculations of the expression, this situation is disabled.

src/optimizer/rule/CollapseProjectRule.cpp
czpmango: Fixed, thx~

jievince (approved):

CPWstatic (commented):
Excellent work.

tests/tck/features/optimizer/CollapseProjectRule.feature
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2021 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License,
+# attached with Common Clause Condition 1.0, found in the LICENSES directory.
+Feature: Push Filter down LeftJoin rule
```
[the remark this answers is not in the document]
czpmango: Fixed, thx~

src/optimizer/rule/CollapseProjectRule.h
```diff
@@ -0,0 +1,37 @@
...
+#define OPTIMIZER_RULE_COLLAPSEPROJECTRULE_H_
+
+#include <memory>
+
+#include "optimizer/OptRule.h"
+
+namespace nebula {
+namespace opt {
+
+class CollapseProjectRule final : public OptRule {
```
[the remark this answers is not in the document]
czpmango: Already added.

yixinglu (approved):

jievince (approved):

Shylock-Hg:
How about the case that the result of collapsed Project may be used by other operator?

jievince:
> How about the case that the result of collapsed Project may be used by other operator?

Is this not checked in `checkDataFlowDeps`?

Shylock-Hg:
> > How about the case that the result of collapsed Project may be used by other operator?
> 
> Is this not checked in `checkDataFlowDeps`?

That seems only check whether data flow same with control flow

PR #1283, opened 2021-07-23
support extended label

jievince:
fix https://github.com/vesoft-inc/nebula-graph/issues/1275
With ` `` `, we can define a `label` with more characters
Maybe we can also support Chinese label via ``


jievince:
Should migrate to nebula repo

PR #1287, opened 2021-07-26
Improve subgraph.

CPWstatic:


Shylock-Hg:
Why change the test case ?

PR #1296, opened 2021-07-29
[DNM] just for ci test

czpmango:
Do Not Merge


czpmango:
Fixed in [https://github.com/vesoft-inc/nebula-graph/pull/1298](https://github.com/vesoft-inc/nebula-graph/pull/1298)

PR #1300, opened 2021-08-02
Optimize/release condition loop branch

Shylock-Hg:
Release branch body when Loop exit.
Release branch which don't execute in Select.


Shylock-Hg:
Move to https://github.com/vesoft-inc/nebula/pull/2518

PR #1306, opened 2021-08-04
Multiple pattern parser

czpmango:


czpmango:
Temporarily close it, there will be a better scheme for related features later.

PR #1311, opened 2021-08-06
push filter down inner join

czpmango:


czpmango:
Previously interrupted and implemented incomplete PR.

PR #1341, opened 2021-08-23
Add a failed piped fetch clause test case

kikimo:
Add a failed piped fetch clause test case.


HarrisChu:
please transfer the pr to 
https://github.com/vesoft-inc/nebula

PR #1349, opened 2021-08-31
Update README.md

Sophie-Xie:
1. update for move repo.
2. fix the logo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="https://github.com/vesoft-inc/nebula/raw/master/docs/logo.png"/>
+  <img src="https://raw.githubusercontent.com/vesoft-inc/nebula/v1-head/docs/logo.png"/>
   <br> English | <a href="README-CN.md">中文</a>
   <br>A distributed, scalable, lightning-fast graph database<br>
 </p>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,8 @@ Compared with other graph database solutions, **Nebula Graph** has the following
 
 The Nebula Graph team is glad to announce the Nebula Graph 2.0.0 GA. This is a brand new start with several important updates, and more of them are on the way. Welcome to join us.
 
+After v2.5.0, we use [Nebula repo](https://github.com/vesoft-inc/nebula).
```
randomJoe211, line 36: Do we need to update the paragraph above this one and change "Nebula Graph 2.0.0 GA" to "Nebula Graph 2.5.0"?

jamieliu1023: Yes I think so if we want to keep the Notice of Release section. However, as a rule of thumb, I would not suggest we put such kind of notice in README because the versions are constantly evolving. It is just annoying that you have to update the version number every time there is a new release. People know where to find the latest version of the product. It just should NOT be a necessary part of a README file IMHO.

Sophie-Xie: Yes, I deleted this description in the readme of the nebula repo.

randomJoe211 (approved):

yixinglu (approved):

PR #1361, opened 2022-08-17
doc: announce the latest repo

TommyLemon:
doc: announce the latest repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,71 +16,7 @@
   <br>
 </p>
 
-# What is Nebula Graph?
-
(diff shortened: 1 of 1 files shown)


wey-gu (commented):
I prefer to add a highlight line in the very first line with notice sign rather than removing everything.

What do you think?

And it's good to highlight this, btw, great catch 👍🏻

TommyLemon:
It does make sense.
I created a new pr:
https://github.com/vesoft-inc/nebula-graph/pull/1362

PR #1362, opened 2022-08-17
doc: announce the latest repo

TommyLemon:


TommyLemon:
I noticed that the Chinese README has already been updated with the announcement.
image

As well as the English README after this commit just now:
https://github.com/vesoft-inc/nebula-graph/commit/74ebed9527153579cd166ead86ebe9cf6b04835e

image

wey-gu (approved):

wey-gu:
> I noticed that the Chinese README has already been updated with the announcement. image

Aha, yes, as I recalled we added a notice, but came to see it's not there ;), it turned out only CN README was added.

